PR #73, opened 2026-04-10
Add TanStack Start server adapter

AntPerez69367:
Hello, this PR is to add an adapter for Tanstack Start. 

It adds:
- New `@neondatabase/auth/start` export — client-side `createAuthClient()` for same-origin proxy auth
- New `@neondatabase/auth/start/server` export — `createNeonAuth()` with handler, middleware, and all Better Auth server methods
- 19 tests covering deferred config resolution, validation, and return structure
- Re-exports via `@neondatabase/neon-js/auth/start` and `@neondatabase/neon-js/auth/start/server`

I based the implementation off the nextjs adapter however unlike the Next.js adapter which takes config directly, `createNeonAuth` accepts a `() => NeonAuthConfig` callback. This is required because TanStack Start's isomorphic execution model means all module-level code runs on both client and server. This approach ensures that the callback is only invoked server-side on first use, keeping secrets out of the client bundle.

```typescript
// Safe — callback deferred until server-side execution
export const auth = createNeonAuth(() => ({
  baseUrl: process.env.NEON_AUTH_URL!,
  cookies: { secret: process.env.NEON_AUTH_COOKIE_SECRET! },
}));
```

Additionally, The adapter uses `await import("@tanstack/react-start/server")` instead of a static import to get `getRequest` and       
`setCookie`. TanStack Start's `@tanstack/react-start/server` entry point exports its entire server runtime, including internal modules that depend on Vite virtual modules only available during SSR. A static import causes Vite's dev server to walk this full dependency graph during module analysis, failing on the unresolvable virtual modules. The dynamic import defers this to runtime, when the code executes inside a server function where everything resolves correctly.


packages/auth/src/start/server/index.ts
```diff
@@ -0,0 +1,141 @@
...
+ *   return session?.user ?? null;
+ * });
+ * ```
+ */
+export function createNeonAuth(getConfig: () => NeonAuthConfig) {
+	let cachedServer: NeonAuthServer | null = null;
+
+	function getServer(): NeonAuthServer {
+		if (!cachedServer) {
+			const config = getConfig();
```
Shridhad: Seems `getConfig()` is being called twice - a) `getServer()` and b) `getValidatedConfig()` -- can produce unknown side-effects.

Centralize to one g`etValidatedConfig()` shared by both paths.

packages/auth/package.json
```diff
@@ -81,6 +81,14 @@
     },
     "./ui/tailwind": {
       "default": "./dist/ui/tailwind.css"
+    },
+    "./start": {
```
Shridhad: The namespace `@neondatabase/auth/start` sounds ambiguous and could collide with future work. Lets be tanstack specific and change the namespace to `@neondatabase/auth/tanstack/start`?

Shridhad:
Hey @AntPerez69367 — thanks for putting this together! We've been wanting TanStack Start support and this is a solid start. The dynamic `import()` for `@tanstack/react-start/server`, and the `createMiddleware` approach are all well thought out.

Full disclosure: we've been working on a parallel implementation on a `feat/tanstack-start` branch, but couldn't finalize. There's a lot of overlap, and rather than duplicate effort we'd love to converge. Your middleware pattern is something we don't have yet and want to adopt. A few things we'd want to align on though:

---

**The big one — reusing the core Proxy**

The SDK has a `createApiProxy()` inside `server/client-factory.ts` that does a lot of heavy lifting: endpoint routing via `API_ENDPOINTS`, cookie forwarding through `fetchWithAuth`, session data cookie minting (the thing that makes cached `getSession()` fast), and a `getSession` override that checks the signed cookie cache before hitting the network.

Critically, that inner Proxy has a `set` trap — it's designed so framework adapters can do exactly `(server as NeonAuth).handler = myHandler` and it just works. The Next.js adapter does this and avoids needing an outer Proxy wrapper entirely.

This PR wraps the inner Proxy in a new `Proxy({} as ...)` with a `get` trap that casts through `Record<string, unknown>`, which erases the type information TypeScript had from `NeonAuthServer`. The `{} as NeonAuthServer & { handler; middleware }` assertion is a fiction the compiler trusts but nothing enforces at runtime. String-based checks like `if (prop === "handler")` also aren't type-checked — rename the property and TS won't catch it. There's also no `has` trap, so `'handler' in auth` would return `false`.

The fix is straightforward — attach `handler` and `middleware` onto the inner proxy via its `set` trap inside `getServer()`, same as Next.js does. The outer proxy (needed for deferred init) then just forwards without needing special cases.

---

**Double config resolution**

Right now `handler.ts` has its own `getValidatedConfig()` with its own cache, separate from the one in `index.ts`. So config gets resolved and validated twice, independently. If we centralize to one resolution path (resolve once in `getServer()`, pass the result to both handler and server), we avoid the duplication and the risk of the two caches drifting.

Related: in both caches, config is assigned before validation runs — so if `validateCookieConfig` throws, the invalid config stays cached and the next call silently skips re-validation. Quick fix: validate a local, then assign.

---

**Smaller things:**

- **Export paths** — `@neondatabase/auth/start` is a bit generic. We went with `@neondatabase/auth/tanstack/start` on our branch to avoid future collisions. Would be good to align.

- **`_splat` param** — `ctx.params._splat` can be undefined for catch-all routes. Needs a `?? ''` fallback.

- **Cookie reads** — `extractNeonAuthCookies(request.headers)` reads raw headers, but cookies set via TanStack's APIs in the same request cycle might not show up there. Worth checking if `getCookie()` from the Start server API is more reliable.

- **Static import in middleware.ts** — You correctly identified the Vite virtual module issue for other imports and used dynamic `import()`, but `createMiddleware` is imported statically. Same class of problem could surface.

- **`@ts-expect-error` for baseUrl** — Works but fragile. A typed overload or explicit same-origin mode would be cleaner.

---

**What we want to take from this PR**

The `createMiddleware().server()` + `context.auth` pattern is great — it's the idiomatic way to do auth in Start server functions and we don't have it yet. We'd like to combine it with our `protectRoute()` helper (which hooks into `beforeLoad` + `redirect()` for route-level guards) so apps get both function-level and route-level auth.

---

Happy to sync on how to merge these approaches — we can iterate here or collaborate on the branch directly. Either way, appreciate you taking this on 🙏

packages/auth/src/start/server/index.ts
```diff
@@ -0,0 +1,141 @@
...
+		return cachedServer;
+	}
+
+	const handler = createHandlerFromConfig(getConfig);
+	const middleware = createMiddlewareFromServer(getServer);
+
+	return new Proxy({} as NeonAuthServer & {
+		handler: ReturnType<typeof createHandlerFromConfig>;
+		middleware: ReturnType<typeof createMiddlewareFromServer>;
+	}, {
```
Shridhad: The new outer Proxy wrapper might bring some unknown issues at runtime. Its better to override by setting `handler` and `middleware` on existing proxy from `createAuthServerInternal`. Approach next.js adapter follows.

AntPerez69367: I actually don't like the Proxy wrapper here and have been working on an alternative solution where we do not need a Proxy at all. I first looked into TanStacks' `createIsomorphicFn` to provide a server and client implementation that differ depending on where they are called, but I think the better route is just to mark the tanstack/start/server implementation as server-only and eliminate any client side running entirely.

AntPerez69367:
Addressed all comments. The main change is removing the outer Proxy wrapper.
                                                                                                     
  TanStack Start's isomorphic execution model evaluates                                    
  createServerFn().middleware([auth.middleware()]) on both client and server. The original Proxy
  approach caused getServer() to run on the client — triggering config validation with missing
  secrets and pulling in node:async_hooks.                                                         
                                                                                                   
  I explored two alternatives:

  - createIsomorphicFn — provides separate server/client implementations at compile time. This works,
   but the client stub still needs to provide matching function signatures (at minimum middleware()),
   and static imports in the file still pull server-only modules into the client bundle.
  - server-only marker — adding import '@tanstack/react-start/server-only' tells TanStack's Vite
  plugin to stub the entire module on the client build. This is the approach I went with — it
  eliminates the need for createIsomorphicFn, client stubs, and the outer Proxy entirely. The server
  implementation now uses the inner proxy's set trap directly to attach
  handler/middleware/protectRoute, matching the pattern used by the Next.js adapter.

I think the approach should be ok. The server will log a warning in development environments that client side code was blocked from accessing server-only but I think that should be ok. Would love to hear your opinon.

packages/auth/src/neon-auth.ts
```diff
@@ -10,6 +10,15 @@ import type {
...
 
+/**
+ * Sentinel for same-origin auth requests (no base URL needed).
+ * Used by framework adapters where the client calls a local `/api/auth` proxy.
+ *
+ * Must be imported from this module. Creating a new `Symbol('same-origin')`
+ * will NOT match, since every `Symbol()` call produces a unique value.
+ */
+export const SAME_ORIGIN = Symbol("same-origin");
+
```
AntPerez69367, line 21: This might be overkill for a same origin mode. I can replace it with a null check too but this felt more intentional.

bun.lock
AntPerez69367, line 1: The bun.lock may be wrong. I dont have a databricks proxy so when i ran bun install it got removed

AntPerez69367:
It may just be better to go the `createIsomorphicFn` approach and stub out the middleware.

PR #62, opened 2026-03-25
ci: two-stage secure release pipeline (Stage 1: prepare-release)

Shridhad:
## Summary
- Migrates CI/CD from bun to pnpm with SHA-pinned action-setup
- Adds JFrog OIDC composite action for secure dependency fetching
- Hardens all workflows (harden-runner, ignore-scripts, min-release-age)
- Fixes peer dependency issues, cache key gaps, and undeclared deps (ACE review)
- Adds release runbook and CI failure recovery guidance

## Related PRs
- **Stage 2**: databricks/secure-public-registry-releases-eng#39 — secure release workflow (install, audit, build, scan, publish via OIDC)

## Test plan
- [ ] CI passes (build, lint, typecheck, unit tests, E2E)
- [ ] Verify pnpm install --frozen-lockfile works cleanly
- [ ] Verify JFrog OIDC action authenticates in CI
- [ ] Run Stage 2 with DRY_RUN=true to validate full pipeline

#0000

PR #77, opened 2026-04-20
chore: fix #62 regressions (scoped)

thekauer:
## Problem

Three regressions introduced by #62, plus a CI-runner fix and a post-review pin tacked on:

1. `next` peer range loosened from `>=16.0.6` to `>=16.0.0` in `packages/auth/package.json`. Code relies on features introduced in 16.0.6+.
2. 33 runtime dependencies in `packages/auth-ui/package.json` pinned to exact versions. Pinning runtime deps in a published library is an anti-pattern — forces duplicate copies into consumer bundles, causes React hook identity errors, defeats tree-shaking.
3. `auth-ui/tsdown.config.ts` left the 5 captcha SDKs as external references, which forced every downstream consumer to stub them via `resolve.alias: { '...': false }` to build. Fix: bundle them into auth-ui's dist via `noExternal`, and switch to `platform: 'browser'` + `fixedExtension: true` so the emitted ESM works in Next.js / Turbopack without a `createRequire` shim. Full rationale in "Why `noExternal`..." below.
4. `.github/workflows/claude*.yml` ran on public `ubuntu-latest` runners instead of the `neondatabase-protected-runner-group` used elsewhere in the repo.

## Summary of changes

- [x] Revert `next` peer range to `>=16.0.6` in `packages/auth/package.json`
- [x] Restore caret ranges on the 33 runtime deps in `packages/auth-ui/package.json`
- [x] Bundle captcha SDKs into `auth-ui/dist` via explicit `noExternal` in `tsdown.config.ts` (not moved to peer deps — see rationale below)
- [x] Switch `auth-ui` tsdown to `platform: 'browser'` + `fixedExtension: true` to emit clean ESM (no `createRequire` shim) while preserving `.mjs` output
- [x] Also `noExternal: [/^@better-auth\//, ...]` so `@better-auth/*` subpackages are inlined as they were pre-#62
- [x] Stop inlining `@neondatabase/auth-ui` into `@neondatabase/auth` (drop `clientPackages` from `preserveDirectives`) to fix consumer transitive-dep resolution
- [x] `packages/internal`: drop unused `composite: true` and disable dts generation as a temporary rolldown-plugin-dts workaround
- [x] Record rolldown platform bindings in `pnpm-lock.yaml` via `supportedArchitectures` in root `package.json`
- [x] Move `.github/workflows/claude*.yml` onto `neondatabase-protected-runner-group` (`linux-ubuntu-latest` label)
- [x] Pin `@marsidev/react-turnstile` to `^1.1.0` per Shridhar's review

## Note: captcha-stub crash is separate concern

Consumer apps that stub captcha SDKs via webpack `resolve.alias = false` (because they don't use captcha) will still crash at render — `@daveyplate/better-auth-ui`'s `useCaptcha` calls `useGoogleReCaptcha` unconditionally at module scope. Fix that crash at the consumer layer by installing the captcha SDKs they use, OR file an upstream fix with daveyplate for lazy captcha imports. This PR doesn't address that crash, scope-limited to the #62 regressions.

## Why `noExternal` + `platform: 'browser'` + `fixedExtension: true`

**Why captchas are bundled (not external):**
- tsdown's default is to externalize everything in `dependencies`. Before these fixes, even with captcha SDKs in `dependencies`, they stayed external — daveyplate's module-scope imports leaked into consumer bundles as external references, forcing every downstream app to ship the same `resolve.alias: { '@wojtekmaj/react-recaptcha-v3': false, ... }` stub dance.
- The user's directive was "no stubs in consumer apps." The minimum-maintenance path (daveyplate has no dynamic provider injection point) is to bundle the captcha SDKs directly into auth-ui's dist.
- Added explicit `noExternal: [/^@daveyplate\/better-auth-ui/, '@wojtekmaj/react-recaptcha-v3', 'react-google-recaptcha', '@hcaptcha/react-hcaptcha', '@captchafox/react', '@marsidev/react-turnstile']` to override tsdown's externalization. Bundle size went from 395 kB → 514 kB (+119 kB / +29 kB gzip).
- Note: captcha is currently dead-code inherited from daveyplate — we don't use it anywhere in neon-auth backend or neon-js wiring. Bundling absorbs the 120 kB in exchange for zero ongoing maintenance.

**Why `platform: 'browser'`:**
- `react-google-recaptcha@3.1.0` ships both CJS (`lib/*.js`) and ESM (`lib/esm/*.js`). Rolldown's default `mainFields` is `['main']` → picked CJS → wrapped it in a CJS-interop stub → emitted `import { createRequire } from "node:module"` at the top of auth-ui's dist.
- Next.js / Turbopack rejects external `node:module` in browser context → build failed with `Code generation for chunk item errored: the chunking context (unknown) does not support external modules (request: node:module)`.
- `platform: 'browser'` makes rolldown's default `mainFields` → `['browser', 'module', 'main']`, ESM-first. The clean ESM path doesn't need the createRequire shim.
- (We also tried `resolve: { mainFields: [...] }` directly — tsdown silently ignores that option, only `resolve.alias` is forwarded.)

**Why `fixedExtension: true`:**
- `fixedExtension` forces `.mjs` output (and `.cjs` for CJS) regardless of `package.json "type"`.
- Default is `true` only when `platform: 'node'`. Flipping to `platform: 'browser'` flipped `fixedExtension` default to `false` → output would be `.js`.
- `packages/auth-ui/package.json` hardcodes `./dist/index.mjs` in `main` + `exports` — without `fixedExtension: true`, exports point at non-existent files and consumer imports break.
- Re-added explicitly to keep `.mjs` output matching existing exports.


packages/auth-ui/tsdown.config.ts
```diff
@@ -11,9 +11,19 @@ export default defineConfig(
...
-    noExternal: [/^@daveyplate\/better-auth-ui/],
+    noExternal: [
+      /^@daveyplate\/better-auth-ui/,
+      /^@better-auth\//,
+      '@wojtekmaj/react-recaptcha-v3',
+      'react-google-recaptcha',
+      '@hcaptcha/react-hcaptcha',
+      '@captchafox/react',
+      '@marsidev/react-turnstile',
+    ],
```
Shridhad, line 26: this will include these libraries into bundle right?  

They are also mentioned in `dependencies` that means every consumer will end up installing these libraries but unused. Should we move them to `devDependencies`?  - or will that cause any other issues?

packages/auth-ui/package.json
```diff
@@ -81,41 +81,41 @@
...
-    "@radix-ui/react-select": "2.2.6",
-    "@radix-ui/react-separator": "1.1.8",
-    "@radix-ui/react-slot": "1.2.4",
-    "@radix-ui/react-tabs": "1.1.13",
-    "@radix-ui/react-use-callback-ref": "1.1.1",
-    "@radix-ui/react-use-layout-effect": "1.1.1",
-    "@wojtekmaj/react-recaptcha-v3": "0.1.4",
+    "@hcaptcha/react-hcaptcha": "^1.12.1",
+    "@hookform/resolvers": "^5.2.2",
+    "@marsidev/react-turnstile": ">=1.1.0",
```
Shridhad: Is this intentional? This will allow any major version. should it be `^1.1.0`?

packages/auth/package.json
```diff
@@ -120,7 +120,7 @@
   "peerDependencies": {
     "react": ">=18.0.0",
     "react-dom": ">=18.0.0",
-    "next": ">=16.0.0"
+    "next": ">=16.0.6"
```
Shridhad: We should support all version from `16.0.0`

Shridhad (commented):
I believe, the PR description is bit outdated

Shridhad (approved):

PR #78, opened 2026-04-20
fix(auth): normalize adapter errors to AuthApiError (.status + .code)

thekauer:
## Problem

`llms.txt:199-202` promises consumers `AuthApiError` with `.status` and `.code`. The Supabase adapter delivers it via `normalizeBetterAuthError` in `packages/auth/src/core/better-auth-helpers.ts`. But the DEFAULT Better Auth adapter path doesn't:

- `packages/auth/src/core/adapter-core.ts:83-88` (force-fetch)
- `packages/auth/src/core/adapter-core.ts:122-128` (main customFetchImpl)
- `packages/auth/src/server/client-factory.ts:111-119` (server-side error return)

All three threw/returned an untyped `Error` with monkey-patched `.status`, and dropped `errorBody.code` from the upstream response. Result: consumers couldn't `instanceof AuthApiError` to discriminate errors on the default adapter path. They've been writing hand-rolled cast helpers to bridge (e.g. `examples/neon-auth-orgs-example/src/lib/auth/errors.ts:31-42`).

## Fix

Route all 3 sites through the existing `normalizeBetterAuthError` helper (no new surface area — the helper already accepts `{ status, statusText, message?, code? }` shaped inputs). Consumers now get a consistent `AuthApiError` (with `.status` + `.code`) regardless of adapter path.

## Summary of changes

- [x] Route `adapter-core.ts` force-fetch + main throw sites through `normalizeBetterAuthError`
- [x] Route `client-factory.ts` error branch through `normalizeBetterAuthError`; `error` field is now an `AuthApiError` instance
- [x] Tests for the default adapter path covering: `.status` + `.code` present, consumer `instanceof` narrowing, typed `.code` when upstream lacks it, 2xx returns unchanged, backward-compat `.status` as `number`

## Consumer impact

Backward compatible in shape — consumers who were doing `(err as { status?: number })?.status` or similar still work (AuthApiError extends Error and exposes `.status: number`), but now `err instanceof AuthApiError` succeeds and they can read `.code` too. Example `phone-sign-in-form.tsx:55` cast becomes unnecessary. `toNeonAuthError` helper in orgs-example becomes unnecessary.

## Verification

- `pnpm -F @neondatabase/auth test:ci` — 121/121 pass (up from 115; 6 new tests)
- `pnpm -F @neondatabase/auth build` — passes (tsdown)
- Lint clean on all touched files
- Manual check: a consumer `catch (err) { if (err instanceof AuthApiError) { err.status; err.code; } }` now compiles without casts

### DIFF
--- a/packages/auth/src/server/client-factory.test.ts
+++ b/packages/auth/src/server/client-factory.test.ts
@@ -1,0 +1,173 @@
+import { describe, test, expect, beforeEach, afterEach, vi } from 'vitest';
+import { createAuthServerInternal } from './client-factory';
+import type { RequestContext } from './request-context';
+import { AuthApiError, AuthError } from '@/adapters/supabase/auth-interface';
+
+const TEST_SECRET = 'test-secret-at-least-32-characters-long!';
+const TEST_BASE_URL = 'https://auth.example.com';
+
+function makeContext(): RequestContext {
+  return {
+    getCookies: () => '',
+    setCookie: () => {},
+    getHeader: () => null,
+    getOrigin: () => 'https://app.example.com',
+    getFramework: () => 'test',
+  };
+}
+
+describe('createAuthServerInternal error branch', () => {
+  const originalFetch = globalThis.fetch;
+  let fetchMock: ReturnType<typeof vi.fn>;
+
+  beforeEach(() => {
+    fetchMock = vi.fn();
+    globalThis.fetch = fetchMock as unknown as typeof globalThis.fetch;
+  });
+
+  afterEach(() => {
+    globalThis.fetch = originalFetch;
+  });
+
+  test('returns { data: null, error: <AuthApiError> } with .status and .code on non-2xx', async () => {
+    fetchMock.mockResolvedValueOnce(
+      Response.json(
+        {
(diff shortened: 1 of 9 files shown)


packages/auth/src/server/client-factory.test.ts
```diff
@@ -0,0 +1,243 @@
...
+import type { RequestContext } from './request-context';
+
+const TEST_SECRET = 'test-secret-at-least-32-characters-long!';
+const TEST_BASE_URL = 'https://auth.example.com';
+
+// Server-return error shape contract: a plain serializable object. It is NOT
+// an `AuthApiError` instance — consumers on the server path depend on
+// `JSON.stringify(error)` and `{...error}` working, and neither works on an
+// `Error` subclass (whose `message` is non-enumerable).
+interface ServerErrorPojo {
```
Shridhad: just `ServerError`?

Shridhad (approved):

packages/auth/src/server/client-factory.test.ts
thekauer: Done in 92415a6.

PR #80, opened 2026-04-22
ci: PR-based Stage 1 release flow + automated tag/handoff

thekauer:
## Summary

Fixes the broken `prepare-release.yml` Stage 1 (it tried to `git push --follow-tags` to protected `main` and failed, leaving orphan tags). Replaces the tail with a PR-based auto-merge flow, adds `post-release.yml` for tag + Stage 2 handoff, and adds `@thekauer` to CODEOWNERS.

**Full design + rationale:** `~/plan/2026-04-22-neon-js-release-flow.md` (local plan file). Key points:

- **No new secrets, no PAT, no GitHub App.** Uses only built-in `github.token` scoped via `permissions:` blocks — the same pattern already used throughout `databricks/secure-public-registry-releases-eng`.
- **3 clicks per release** — dispatch → approve → human-dispatch Stage 2 from the PR comment the bot posts.
- **Branch protection preserved.** Auto-merge waits for CODEOWNER approval + checks. Either @shridhad or @thekauer can approve.

Revert flow (`revert-release.yml`) is intentionally **NOT** in this PR — it ships in a follow-up PR once this foundation lands. The design is documented in the plan (Task 3).

## Changes

- `prepare-release.yml` — actor gate (admin/maintain), version bump on a `release/<pkg>-<run-id>` branch, PR opened with auto-merge enabled. No more push-to-main.
- `post-release.yml` (new) — fires on push to main matching `chore: release @neondatabase/`, tags the squash commit, pushes tags, posts a PR comment + workflow summary with Stage 2 dispatch instructions (CLI one-liner + UI deep link per package).
- `CODEOWNERS` — adds `@thekauer` alongside `@shridhad`. Either can approve.

## Stage 2 side

Stage 2 is tracked on [databricks/secure-public-registry-releases-eng#64](https://github.com/databricks/secure-public-registry-releases-eng/pull/64) — the pnpm + JFrog reorder fix. No Stage 2 changes in this PR.

## Test plan

- [ ] Merge this PR (remove draft) after review
- [ ] Dispatch `prepare-release.yml` with `package=postgrest-js`, `bump=patch`, `ref=main` (leaf cascade, low-risk smoke test)
- [ ] Confirm PR opens with `release` label and auto-merge enabled
- [ ] Approve → PR auto-merges → `post-release.yml` fires → tags pushed → PR comment appears with Stage 2 dispatch links
- [ ] Click the UI deep link / copy the CLI one-liner → dispatch secpub `neon-js.yml` with `dry-run=true` for both `postgrest-js` and `neon-js`
- [ ] Confirm Stage 2 runs complete without publishing (dry-run)

## Follow-ups

- `revert-release.yml` in a separate PR (spec in the plan, Task 3)
- Stage 1 file-header comment block still describes the old "push tags" flow — cosmetic cleanup

This pull request and its description were written by Isaac.


.github/workflows/post-release.yml
```diff
@@ -0,0 +1,156 @@
...
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+
+          TAGS='${{ steps.parse.outputs.tags }}'
+          SHA="${{ github.sha }}"
+
+          for tag in $(echo "$TAGS" | jq -r '.[]'); do
+            if git rev-parse -q --verify "refs/tags/$tag" >/dev/null; then
+              echo "Tag $tag already exists — skipping (idempotent rerun)."
+              continue
```
Shridhad: Should it fail if the tag exists but points to older commit/SHA??

.github/workflows/post-release.yml
```diff
@@ -0,0 +1,156 @@
...
+            fi
+            git tag -a "$tag" "$SHA" -m "Release $tag (commit $SHA)"
+            echo "Created tag $tag → $SHA"
+          done
+          git push origin --tags
```
Shridhad: Lets push only single tag - `git push origin "refs/tags/$tag"`

.github/workflows/revert-release.yml
```diff
@@ -0,0 +1,237 @@
+name: Revert Release
```
Shridhad: I am not sure if we should have revert tag workflow or add logic to override the tag in release flow itself. 

With revert flow, it may end up in conflict if we pushed different commits after tag was created. 🤔

Shridhad (dismissed):

.github/workflows/post-release.yml
thekauer: Fixed in `773b33b`.

Peels the tag via `^{commit}` before comparing — `git tag -a` creates an annotated-tag object whose SHA is distinct from the commit it points to, so a raw compare would always mismatch even on an idempotent rerun. Skip only when the resolved commit matches. On any mismatch, fail loudly with recovery hints and never auto-retarget.

.github/workflows/post-release.yml
thekauer: Fixed in `773b33b`. Now pushes each tag per-iteration with `git push origin "refs/tags/$tag"` inside the loop; dropped the trailing `git push origin --tags` so we don't accidentally push unrelated local tags. Paired with the idempotent `^{commit}` check above, partial-failure reruns are safe.

Shridhad (approved):

PR #82, opened 2026-04-22
fix: Safari OAuth flow broken by Partitioned cookie flag

Shridhad:
## Problem

The upstream Neon Auth server sets cookies (including `session_challange`) with `Partitioned; SameSite=None; Secure`. When the proxy forwards these to the browser as first-party cookies on the user's own domain, Safari stores them but **does not send them on top-level navigations** — which is exactly when the OAuth callback fires.

This broke the Next.js middleware OAuth flow:
1. `POST /api/auth/sign-in/social` → proxy sets `session_challange` with upstream's `Partitioned; SameSite=None`
2. OAuth redirect chain completes → browser navigates to `/callback?neon_auth_session_verifier=...`
3. Middleware calls `needsSessionVerification()` — `session_challange` cookie absent in Safari
4. Middleware skips exchange → no session → redirects to login

## Fix

Strip `Partitioned` and force `SameSite=Lax` on all upstream `Set-Cookie` headers before forwarding them to the browser. These flags made sense when cookies were third-party (from `auth.neon.tech`); once proxied onto the user's own domain they become same-site and `Lax` is both correct and more secure.

Two code paths handle upstream cookies — both are fixed:

**`server/proxy/response.ts`** (API route handler path): previously only sanitized when `domain` was configured, falling through to a verbatim forward otherwise.

**`server/client-factory.ts`** (Next.js server action path): same sanitization applied unconditionally.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,2 +1,2 @@
-# Every PR requires review from @shridhad
-* @shridhad
+# Every PR requires review from @shridhad or @thekauer (Andras)
+* @shridhad @thekauer
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,35 @@
+version: 2
+updates:
+  - package-ecosystem: npm
+    directories:
+      - /
+      - /packages/*
+      - /examples/*
+      - /e2e
+    schedule:
+      interval: weekly
+      day: monday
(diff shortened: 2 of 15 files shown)


thekauer (dismissed):
LGTM!

thekauer (approved):

thekauer (approved):

PR #114, opened 2026-04-30
feat(neon-js): single-URL initialization for createClient (LKB-11187)

ruslantalpa:
## Summary
  - Adds a string-form to `createClient` so apps can initialize with a single base URL instead of two (`auth` + `dataApi`). The existing object form is unchanged.
  - Derivation is configurable: `defaultDeriveNeonUrls` is the built-in for Neon endpoints.
  - Public exports added: `defaultDeriveNeonUrls`, `DeriveNeonUrls`, `CreateClientStringOptions`.

  ## API
  ```ts
  // New: single URL
  const client = createClient<DB>('https://ep-xxx.c-2.us-east-2.aws.neon.build/dbname');

  // New: with overrides
  const client = createClient<DB>(baseUrl, {
    auth: { allowAnonymous: true, adapter: SupabaseAuthAdapter() },
    dataApi: { options: { /* … */ } },
  });

  // Backward compatible: object form
  const client = createClient<DB>({ auth: { url }, dataApi: { url } });
```

  Test plan

  - defaultDeriveNeonUrls unit tests (8 cases incl. trailing slash, custom port, deeper path, http, query/hash rejection, label-count guard).
  - createClient runtime tests with mocked auth factory: derivation, per-call URL overrides, custom deriveUrls, adapter forwarding, allowAnonymous default + flow-through, invalid URL throws,
  object-form backward compat.
  - Type tests for the string form: default adapter, SupabaseAuthAdapter, BetterAuthReactAdapter, Database type-parameter inference.
  - Repo-level typecheck + build clean; package lint clean.
  - Documentation/READMEs (intentionally out of scope — handled separately).

  Refs: LKB-11187. Parent: LKB-11182.

### DIFF
--- a/packages/neon-js/src/client/derive-urls.ts
+++ b/packages/neon-js/src/client/derive-urls.ts
@@ -1,0 +1,48 @@
+/**
+ * Pure function: given a Neon base URL, returns the corresponding
+ * auth and Data API URLs.
+ *
+ * Wrappers (e.g. @databricks/lakebase-js) can supply a different implementation
+ * via the `deriveUrls` option on createClient.
+ */
+export type DeriveNeonUrls = (baseUrl: string) => {
+  auth: string;
+  dataApi: string;
+};
+
+/**
+ * Default URL derivation for Neon endpoints.
+ *
+ * Insert `neonauth` / `apirest` after the first hostname label and append
+ * `/auth` / `/rest/v1` to the path.
+ *
+ * Example:
+ *   https://ep-xxx.c-2.us-east-2.aws.neon.build/dbname
+ *     → auth:    https://ep-xxx.neonauth.c-2.us-east-2.aws.neon.build/dbname/auth
+ *     → dataApi: https://ep-xxx.apirest.c-2.us-east-2.aws.neon.build/dbname/rest/v1
+ */
+export const defaultDeriveNeonUrls: DeriveNeonUrls = (baseUrl) => {
+  const url = new URL(baseUrl);
+
+  const labels = url.hostname.split('.');
+  if (labels.length < 3) {
+    throw new Error(
+      `Invalid Neon base URL: hostname "${url.hostname}" must have at least 3 dot-separated labels (e.g. ep-xxx.region.tld).`
+    );
+  }
+
+  const [first, ...rest] = labels;
+  const authHost = [first, 'neonauth', ...rest].join('.');
+  const dataApiHost = [first, 'apirest', ...rest].join('.');
+
+  const basePath = url.pathname.replace(/\/+$/, '');
+  const authPath = `${basePath}/auth`;
+  const dataApiPath = `${basePath}/rest/v1`;
+
+  const portSuffix = url.port ? `:${url.port}` : '';
+
+  return {
+    auth: `${url.protocol}//${authHost}${portSuffix}${authPath}`,
+    dataApi: `${url.protocol}//${dataApiHost}${portSuffix}${dataApiPath}`,
+  };
+};
--- a/packages/neon-js/src/client/derive-urls.ts
+++ b/packages/neon-js/src/client/derive-urls.ts
@@ -24,6 +24,12 @@ export type DeriveNeonUrls = (baseUrl: string) => {
 export const defaultDeriveNeonUrls: DeriveNeonUrls = (baseUrl) => {
   const url = new URL(baseUrl);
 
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: packages/neon-js/src/client/derive-urls.ts [hunk 1]
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * Example:
+ *   https://ep-xxx.c-2.us-east-2.aws.neon.build/dbname
+ *     → auth:    https://ep-xxx.neonauth.c-2.us-east-2.aws.neon.build/dbname/auth
+ *     → dataApi: https://ep-xxx.apirest.c-2.us-east-2.aws.neon.build/dbname/rest/v1
+ */
+export const defaultDeriveNeonUrls: DeriveNeonUrls = (baseUrl) => {
+  const url = new URL(baseUrl);
+
+  if (url.search !== '' || url.hash !== '') {
```
Shridhad, line 37: Nice guard rejecting `?query`/`#hash`. For symmetry, consider that userinfo (`user:pass@host`) is currently dropped silently — either reject it too or note in the JSDoc that only protocol/host/port/path are used. Low impact.

packages/neon-js/src/client/derive-urls.ts
```diff
@@ -0,0 +1,51 @@
...
+    throw new Error(
+      `Invalid Neon base URL: must not include a query string or hash fragment (got "${baseUrl}").`
+    );
+  }
+
+  const labels = url.hostname.split('.');
+  if (labels.length < 3) {
```
Shridhad, line 50: The `< 3 labels` heuristic also rejects `localhost`/single-host proxies. Reasonable given the `deriveUrls` escape hatch, but worth a sentence in the JSDoc so callers know this constraint is intentional.

packages/neon-js/src/client/derive-urls.ts
```diff
@@ -0,0 +1,51 @@
...
+      `Invalid Neon base URL: hostname "${url.hostname}" must have at least 3 dot-separated labels (e.g. ep-xxx.region.tld).`
+    );
+  }
+
+  const [first, ...rest] = labels;
+  const authHost = [first, 'neonauth', ...rest].join('.');
+  const dataApiHost = [first, 'apirest', ...rest].join('.');
+
+  const basePath = url.pathname.replace(/\/+$/, '');
+  const port = url.port ? `:${url.port}` : '';
```
Shridhad, line 61: A single port is parsed from the base URL and applied to BOTH derived hosts. The repo's own commented local-dev config uses different ports per service (`neonauth:30443` vs `apirest:9443` in examples/react-neon-js/src/client.ts), so the string form can't reproduce a split-port local setup — those users must fall back to the object form or `deriveUrls`. Fine for production (both on 443); worth a one-line doc note.

Shridhad (approved):
Approving — clean, purely additive, and well-tested. The object form is untouched; the string form is layered on via new overloads + `buildConfigFromBaseUrl`. I confirmed `defaultDeriveNeonUrls` produces exactly the auth/Data API URLs used in the example apps (`neonauth`/`apirest` infix + `/auth` and `/rest/v1` suffixes). Build, repo typecheck, lint, and all 39 unit/type tests pass locally.

Left a few non-blocking inline notes (mostly doc/edge-case suggestions). None need to gate the merge.

packages/neon-js/src/client/derive-urls.ts
ruslantalpa: Done in 9611dd8 — now rejects embedded credentials (user:pass@host) explicitly, symmetric with the query/hash guard, with a test covering both user:pass@ and user@forms.

packages/neon-js/src/client/derive-urls.ts
ruslantalpa: Done in 9611dd8 — documented that a single port is applied to both derived hosts, so split-port local setups (e.g. neonauth:30443 vs apirest:9443) need the object form or deriveUrls; both share 443 in production.

packages/neon-js/src/client/derive-urls.ts
ruslantalpa: Done in 9611dd8 — added a JSDoc note that the ≥3-label constraint intentionally rejects localhost/single-host proxies, pointing to the object form / custom deriveUrls as the escape hatch.

PR #134, opened 2026-05-10
feat(auth): Next.js server observability (default warn) and fetch error taxonomy

Shridhad:
## Summary

Adds **observability** for the Neon Auth Next.js server stack (proxy, middleware, Better Auth server `fetch`). By default Neon Auth emits structured **`warn`** / **`error`** logs to `console`. Set **`logLevel: 'silent'`** on `createNeonAuth` / `NeonAuthConfig` to mute completely, or pass **`logger`** / raise **`logLevel`** (`info`, `debug`) for more detail.

Classifies failed upstream **`fetch`** calls into stable JSON **`code`** values (`NETWORK_DNS`, `NETWORK_TIMEOUT`, `NETWORK_TLS`, …) so synthetic **502** responses from the SDK are easier to diagnose when the Neon Auth service shows no elevated 5xx.

## Changes

- `resolveNeonAuthLogging`, `NeonAuthLogger`, `NeonAuthLogLevel` (includes **`silent`**), `NeonAuthActiveLogLevel`, related exports from `@neondatabase/auth` and `@neondatabase/auth/next/server`
- Framework-agnostic wiring in proxy, middleware processor + OAuth exchange, server API client
- **`packages/auth/CHANGELOG.md`** (Unreleased)

## Notes

- **`NeonAuthConfig`** exposes **`logger`** and **`logLevel`** only (no separate **`logging`** boolean); **`logLevel: 'silent'`** disables Neon Auth console output.


packages/auth/src/server/middleware/processor.ts
```diff
@@ -56,6 +58,11 @@ export interface AuthMiddlewareConfig {
 export async function processAuthMiddleware(
 	config: AuthMiddlewareConfig
 ): Promise<MiddlewareResult> {
+	const log = resolveNeonAuthLogging({
```
Shridhad: **Critical: `auth.middleware()` silently ignores user-supplied `logger` and `logLevel`.**

`resolveNeonAuthLogging` here resolves from `config.logger` / `config.logLevel`, but the Next.js wrapper at `packages/auth/src/next/server/middleware.ts` (lines ~43–60) destructures only `{ baseUrl, cookies, loginUrl }` from its input and calls `processAuthMiddleware({ ... })` without forwarding `logger` / `logLevel` — even though `createNeonAuth` does spread them via `neonAuthMiddleware({ ...config, ...middlewareConfig })` (`next/server/index.ts:188`).

Net effect: a user setting `logLevel: 'silent'` or passing a custom `logger` cannot mute or redirect middleware logs — the middleware path always uses the default `warn`-to-`console` sink. This contradicts the PR description, the JSDoc on `createNeonAuth`, the changelog Unreleased entry, and the new `CLAUDE.md` Learned User Preferences line.

**Suggested fix:** since `createNeonAuth` already calls `resolveNeonAuthLogging(config)` once on construction (`next/server/index.ts:129`), extend `AuthMiddlewareConfig` with an optional pre-resolved `log?: ResolvedNeonAuthLogging` and have `neonAuthMiddleware` pass that through. That also subsumes the per-request resolver allocation flagged separately. A `processor.test.ts` case that passes `logger`/`logLevel` and asserts the resolved sink reaches `handleAuthProxyRequest` and `exchangeOAuthToken` would have caught this.

packages/auth/src/server/middleware/processor.ts
```diff
@@ -56,6 +58,11 @@ export interface AuthMiddlewareConfig {
...
+		logger: config.logger,
+		logLevel: config.logLevel,
+	});
```
Shridhad: **Per-request resolver allocation.**

This allocates four new gate closures on every middleware invocation. `createNeonAuth` already resolves once (`next/server/index.ts:129`) and `authApiHandler` resolves once per handler factory (`handler.ts:39`) — middleware is the odd one out. Accepting an optional pre-resolved `log?: ResolvedNeonAuthLogging` on `AuthMiddlewareConfig` and having `neonAuthMiddleware` pass it through is the same plumbing fix that resolves the wiring bug above, with one less allocation per request as a bonus.

Shridhad (commented):
Thanks for shipping this — the centralized logging shape (`resolveNeonAuthLogging` with `'silent'` short-circuit + `console` fallback) and the `classifyFetchFailure` cause-chain / AggregateError handling are both clean foundations, and the new `NeonAuthNetworkErrorCode` is the right thing to expose publicly. Posting one critical wiring issue plus a set of important and nice-to-have observations inline. Happy to elaborate or pair on any of them.

packages/auth/src/server/middleware/processor.ts
```diff
@@ -120,6 +128,7 @@ export async function processAuthMiddleware(
 			sessionDataTtl,
 			domain,
 			sameSite,
+			log,
 		});
```
Shridhad, line 136: **(Suggestion, not introduced by this PR.)** Now that you have a `log` in scope, the silent JSON parse swallow two lines below — `await sessionResponse.json().catch(() => null)` (line ~136) — is a good opportunity to surface diagnosis info on `debug`/`warn` when upstream returns a malformed body. Today it silently presents the user as logged-out:

```ts
const data = await sessionResponse.json().catch((err) => {
  log?.debug('[neon-auth] Failed to parse session response JSON', {
    component: 'middleware',
    detail: err instanceof Error ? err.message : String(err),
  });
  return null;
});
```

packages/auth/src/server/network-error.ts
```diff
@@ -0,0 +1,206 @@
...
+
+function readErrnoCode(node: unknown): string | undefined {
+	if (!node || typeof node !== 'object') return undefined;
+	const code = (node as NodeJS.ErrnoException).code;
+	return typeof code === 'string' ? code : undefined;
+}
+
+function isAbortError(node: unknown): boolean {
+	return node instanceof Error && node.name === 'AbortError';
+}
```
Shridhad, line 62: **`isAbortError` may miss aborts on newer runtimes.**

Several runtimes (Node ≥ 18 in some configurations, browsers, Edge) throw a `DOMException` (not an `Error` subclass) for fetch aborts. Such aborts will fall through here and either match `NETWORK_RESET` via `ECONNABORTED` later in the chain, or be classified as `internal`.

```suggestion
function isAbortError(node: unknown): boolean {
	if (node instanceof Error && node.name === 'AbortError') return true;
	if (
		typeof node === 'object' &&
		node !== null &&
		'name' in node &&
		(node as { name?: unknown }).name === 'AbortError'
	) {
		return true;
	}
	return false;
}
```

A test for `DOMException` (`name === 'AbortError'`) would lock this in.

packages/auth/src/server/network-error.ts
```diff
@@ -0,0 +1,206 @@
...
+ */
+
+export type NeonAuthNetworkErrorCode =
+	| 'NETWORK_ERROR'
+	| 'NETWORK_DNS'
+	| 'NETWORK_REFUSED'
+	| 'NETWORK_TIMEOUT'
+	| 'NETWORK_TLS'
+	| 'NETWORK_RESET'
+	| 'NETWORK_ABORT';
```
Shridhad: **`NeonAuthNetworkErrorCode` is now part of the observable API.**

These string literals appear as JSON `code` on 502 responses and in the server-API error envelope, so renames are breaking. Two small things would help downstream:

1. A JSDoc note declaring the values semver-stable and enumerating them in one place.
2. An `as const` array for runtime validation / iteration:

```ts
export const NEON_AUTH_NETWORK_ERROR_CODES = [
	'NETWORK_ERROR',
	'NETWORK_DNS',
	'NETWORK_REFUSED',
	'NETWORK_TIMEOUT',
	'NETWORK_TLS',
	'NETWORK_RESET',
	'NETWORK_ABORT',
] as const;
export type NeonAuthNetworkErrorCode =
	typeof NEON_AUTH_NETWORK_ERROR_CODES[number];
```

packages/auth/src/server/network-error.ts
```diff
@@ -0,0 +1,206 @@
...
+	}
+
+	const head = nodes[0];
+	const message =
+		head instanceof Error ? head.message : 'Internal Server Error';
+
+	return {
+		kind: 'internal',
+		detail: message.slice(0, 200),
+		clientMessage: message,
```
Shridhad: **Information disclosure: `kind: 'internal'` returns raw, unsanitized `Error.message` to clients.**

`detail` is clamped to 200 chars but `clientMessage` is the **untruncated** original message. It is forwarded directly into client-visible payloads:

- `proxy/request.ts:95–101` → `Response.json({ error: classified.clientMessage, code: 'INTERNAL_ERROR' }, { status: 500 })` returned to the browser.
- `client-factory.ts:108–116` → returned to RSC / Server Action callers.

Internal exception messages can include file paths, env-laden text, ORM errors, etc. Use a fixed public string for `clientMessage` and keep the real text only in log meta:

```suggestion
	return {
		kind: 'internal',
		detail: message.slice(0, 200),
		clientMessage: 'Internal Server Error',
	};
```

packages/auth/src/server/logger.ts
```diff
@@ -0,0 +1,125 @@
...
+
+/**
+ * Resolved sink used internally after merging defaults and level filtering.
+ */
+export type ResolvedNeonAuthLogging = {
+	error(message: string, meta?: Record<string, unknown>): void;
+	warn(message: string, meta?: Record<string, unknown>): void;
+	info(message: string, meta?: Record<string, unknown>): void;
+	debug(message: string, meta?: Record<string, unknown>): void;
+};
```
Shridhad: **`ResolvedNeonAuthLogging` duplicates `Required<NeonAuthLogger>`.**

`consoleSink: Required<NeonAuthLogger>` and this type are structurally identical. Single source of truth:

```suggestion
export type ResolvedNeonAuthLogging = Required<NeonAuthLogger>;
```

While here: `NeonAuthActiveLogLevel` doesn't appear to have any consumer-facing usage outside this file (grep finds it only in re-exports). Consider not exporting it — every exported symbol is a semver commitment, and you're just adding `Exclude<NeonAuthLogLevel, 'silent'>` if a consumer ever needs the narrowed type. Also: the JSDoc on this type says "used internally" but it is publicly re-exported from `@neondatabase/auth/next/server` — pick a side.

packages/auth/src/server/logger.ts
```diff
@@ -0,0 +1,125 @@
...
+}
+
+export type NeonAuthLoggingInput = {
+	logger?: NeonAuthLogger;
+	/**
+	 * Minimum level for Neon Auth logs. Use **`'silent'`** to disable all Neon Auth `console` output.
+	 * @default 'warn' — emits `error` and `warn` only
+	 */
+	logLevel?: NeonAuthLogLevel;
+};
```
Shridhad, line 94: **`'silent'` ignores a user-supplied `logger`, but the type system allows the contradictory shape.**

`{ logLevel: 'silent', logger: {...} }` type-checks but at runtime `resolveNeonAuthLogging` returns `noopResolved` before merging `logger` (line ~112). Either document this on the JSDoc here (e.g. *"When set to `'silent'`, any provided `logger` is also silenced"*), or use a discriminated union:

```ts
export type NeonAuthLoggingInput =
	| { logLevel: 'silent'; logger?: never }
	| { logLevel?: NeonAuthActiveLogLevel; logger?: NeonAuthLogger };
```

A test that asserts `silent` overrides a custom logger would also be worth adding (current `logger.test.ts` only exercises the default-sink case).

packages/auth/src/server/client-factory.ts
```diff
@@ -65,11 +69,69 @@ export function createAuthServerInternal(
...
+
+      return {
+        data: null,
+        error: {
+          message: classified.clientMessage,
+          status: 500,
+          statusText: 'Internal Server Error',
+          code: 'INTERNAL_ERROR',
+        },
+      };
```
Shridhad: **Behavior change not documented: internal classifications now return a POJO instead of throwing.**

The new `try/catch` around `fetch` is a real improvement for transport errors. But it also catches `kind: 'internal'` and returns `{ data: null, error: { …, status: 500, code: 'INTERNAL_ERROR' } }` to RSC / Server Action callers. Previously, unclassified errors would throw and surface in the framework's error boundary; now they're swallowed into the result envelope.

Either **rethrow** on `kind: 'internal'` (preserves prior behavior; only the *transport* class becomes a structured POJO) or document the change explicitly in the changelog so consumers can update their error boundaries / fallbacks.

packages/auth/src/server/client-factory.ts
```diff
@@ -65,11 +69,69 @@ export function createAuthServerInternal(
...
+        });
+        return {
+          data: null,
+          error: {
+            message: classified.clientMessage,
+            status: 502,
+            statusText: 'Bad Gateway',
+            code: classified.code,
+          },
+        };
```
Shridhad, line 97: **Consider exporting a typed error envelope so consumers can switch on `error.code`.**

Server-API callers (RSC / Server Actions) get `error.code` that is one of `NeonAuthNetworkErrorCode` (transport, here), `'INTERNAL_ERROR'` (internal), or whatever `normalizeBetterAuthError` produces (HTTP). Today there's no exported type for the envelope, so consumers can't narrow it. Adding e.g.:

```ts
export type NeonAuthServerApiError = {
	message: string;
	status: number;
	statusText: string;
	code: NeonAuthNetworkErrorCode | 'INTERNAL_ERROR' | (string & {});
};
```

lets `if (error.code === 'NETWORK_DNS') …` type-narrow downstream without resorting to `as const` casts.

packages/auth/src/server/client-factory.ts
```diff
@@ -116,8 +178,10 @@ export function createAuthServerInternal(
           }
         }
       } catch (error) {
-        // Log error but don't fail the request - session cache is optional
-        console.error('[fetchWithAuth] Failed to mint session data cookie:', error);
+        log?.warn('[neon-auth] Failed to mint session data cookie', {
+          component: 'server-api',
+          detail: error instanceof Error ? error.message : String(error),
+        });
```
Shridhad, line 185: **Stack and `cause` chain are dropped on the migration from `console.error(msg, error)` to `log?.warn(msg, { detail })`.**

Previously `console.error('[fetchWithAuth] Failed to mint session data cookie:', error)` carried the full `Error` (stack + `cause`) — useful for Sentry, OTel, structured loggers. The new shape passes only `error.message`. Since `NeonAuthLogger.meta` is `Record<string, unknown>`, the raw error is welcome:

```suggestion
        log?.warn('[neon-auth] Failed to mint session data cookie', {
          component: 'server-api',
          detail: error instanceof Error ? error.message : String(error),
          err: error,
        });
```

Same applies to the matching site at line ~256 (`Cookie validation error before getSession upstream call`). Worth a JSDoc note on `NeonAuthLogger` that custom loggers may receive non-enumerable `Error` fields under `err`.

**Bonus while you're here**: the comment block just above (lines ~142–150) still says *"force SameSite=Lax for Safari compatibility"*, but the code now uses `effectiveSameSite` (default `strict`), and `proxy/response.ts` was updated for `0.4.0-beta` to read *"Apply configured SameSite (default strict)"*. The Lax wording here is stale rot — please drop it in the same touch:

```ts
// Mirror sanitization from prepareResponseHeaders (response.ts):
// strip Partitioned and apply configured SameSite (default `strict`).
// Always override domain: use local config if set, otherwise strip any
// upstream Domain attribute to avoid leaking the auth server's domain.
```

packages/auth/src/server/proxy/request.ts
```diff
@@ -8,45 +10,97 @@ const PROXY_HEADERS = ['user-agent', 'authorization', 'referer', 'content-type']
...
+				host,
+			});
+		} else {
+			log?.warn('[neon-auth] Upstream HTTP error', {
+				component: 'proxy',
+				proxyPath: path,
+				status: response.status,
+				statusText: response.statusText,
+				host,
+			});
```
Shridhad, line 62: **Consider differentiating routine 4xx from 5xx for log severity.**

Logging `warn` for *any* non-`ok` upstream response will produce noise on routine `401`/`403` from `/get-session` (expired sessions, anonymous users hitting protected endpoints, etc.). Suggest `info`/`debug` for 4xx (expected auth failures) and `warn` only for 5xx / transport:

```ts
} else if (response.status >= 500) {
	log?.warn('[neon-auth] Upstream HTTP error', { … });
} else {
	log?.info('[neon-auth] Upstream HTTP non-2xx', { … });
}
```

Same pattern in `client-factory.ts:127–134`.

packages/auth/src/server/proxy/request.ts
```diff
@@ -56,46 +110,50 @@ export const handleAuthRequest = async (baseUrl: string, request: Request, path:
...
-    return request.text();
-  }
+	if (request.body) {
+		return request.text();
+	}
 
-  return;
-}
+	return;
+};
```
Shridhad: **Style nits while touching this file:**

1. File ends without a trailing newline (`\ No newline at end of file` in the diff).
2. `proxy/request.ts` and `middleware/processor.ts` use **tabs** while neighbouring files in the package (`client-factory.ts`, `next/server/index.ts`) use **2-space indent**. Running prettier on the touched files would normalize both.

packages/auth/src/next/server/index.ts
```diff
@@ -40,6 +51,7 @@ export {
  * @param config.cookies.secret - Secret for signing session cookies (minimum 32 characters)
  * @param config.cookies.sessionDataTtl - Optional TTL for session cache in seconds (default: 300)
  * @param config.cookies.domain - Optional cookie domain (default: current domain)
+ * @param config.logLevel - Set `'silent'` to disable Neon Auth server console logs (default: `warn`)
```
Shridhad: **Doc gaps for the new public surface.**

A few small but discoverable items:

1. This JSDoc adds `@param config.logLevel` but is missing a paired `@param config.logger`. Suggested:
   ```
   * @param config.logger - Optional structured logger; omitted methods fall back to `console` (see {@link NeonAuthLogger})
   ```
2. `exchangeOAuthToken` (`server/middleware/oauth.ts:54–62`) — the new `log` param has no `@param` entry.
3. `NeonAuthLogLevel` opening line (`server/logger.ts:5`) reads *"Levels that emit output"* but the union now includes `'silent'`, which doesn't emit. Consider: *"Supported Neon Auth log levels (including `'silent'` to disable all SDK console output)"*.
4. `NeonAuthNetworkErrorCode` and `ClassifiedFetchFailure` are publicly exported with no JSDoc — users need a stable, discoverable enumeration of the codes they can switch on (see the separate inline on `network-error.ts:12`).
5. `ResolvedNeonAuthLogging`'s JSDoc says *"used internally"* but is publicly re-exported here — drop the export or rewrite the doc.

packages/auth/CHANGELOG.md
```diff
@@ -8,6 +8,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **Next.js server observability (opt-out defaults)**: `createNeonAuth` / `NeonAuthConfig` support **`logger`** and **`logLevel`**. Defaults to **`warn`** on `console` (structured `error` / `warn` from the server proxy, middleware, and Better Auth server `fetch`). Set **`logLevel: 'silent'`** to disable Neon Auth `console` output; use **`info`** / **`debug`** or a custom **`logger`** for more detail.
+- **Logging types**: `NeonAuthLogLevel` includes **`'silent'`**; **`NeonAuthActiveLogLevel`** (levels that actually emit) is exported from `@neondatabase/auth` and `@neondatabase/auth/next/server` alongside `resolveNeonAuthLogging`, `NeonAuthLogger`, `ResolvedNeonAuthLogging`, and `NeonAuthLoggingInput`.
+- **Upstream fetch error taxonomy**: Failed upstream `fetch` calls are classified into stable JSON **`code`** values (`NETWORK_DNS`, `NETWORK_TIMEOUT`, `NETWORK_TLS`, etc.) instead of only `NETWORK_ERROR`, including on synthetic **502** responses from the SDK proxy.
```
Shridhad: **Worth listing the full set of stable codes (or linking to `NeonAuthNetworkErrorCode`).**

The bullet enumerates `NETWORK_DNS`, `NETWORK_TIMEOUT`, `NETWORK_TLS` as examples but the full set is `NETWORK_ERROR | NETWORK_DNS | NETWORK_REFUSED | NETWORK_TIMEOUT | NETWORK_TLS | NETWORK_RESET | NETWORK_ABORT`. Anyone writing `switch (error.code) { … }` will want all seven listed in one place — either expand the bullet or append *"see `NeonAuthNetworkErrorCode` for the full union"*.

packages/auth/src/server/middleware/processor.ts
Shridhad: Done in latest push: `neonAuthMiddleware` forwards `logger`, `logLevel`, and the pre-resolved `log` from `createNeonAuth`; `processAuthMiddleware` uses `config.log ?? resolveNeonAuthLogging(...)`. Added `processor.test.ts` asserting `handleAuthProxyRequest` receives the forwarded sink.

packages/auth/src/server/middleware/processor.ts
Shridhad: Same fix — optional `log` on the middleware path avoids per-call `resolveNeonAuthLogging` when using `createNeonAuth`.

packages/auth/src/server/middleware/processor.ts
Shridhad: Added `log.debug` when `sessionResponse.json()` fails in middleware so malformed upstream bodies are visible without failing open silently.

packages/auth/src/server/network-error.ts
Shridhad: Applied the suggested `isAbortError` shape + `DOMException` test.

packages/auth/src/server/network-error.ts
Shridhad: Added `NEON_AUTH_NETWORK_ERROR_CODES` `as const`, derived `NeonAuthNetworkErrorCode`, semver stability note in module JSDoc.

packages/auth/src/server/network-error.ts
Shridhad: `kind: internal` now uses a fixed public `clientMessage` (`Internal Server Error`); truncated detail stays server-side / logs only.

packages/auth/src/server/logger.ts
Shridhad: `ResolvedNeonAuthLogging` is now `Required<NeonAuthLogger>`. Dropped `NeonAuthActiveLogLevel` from public exports and clarified `ResolvedNeonAuthLogging` docs.

packages/auth/src/server/logger.ts
Shridhad: Documented that `logLevel: silent` ignores any `logger`; added `logger.test.ts` asserting silent wins.

packages/auth/src/server/client-factory.ts
Shridhad: `fetchWithAuth` now rethrows after logging when classification is non-transport (preserves prior boundary behavior). CHANGELOG documents this.

packages/auth/src/server/client-factory.ts
Shridhad: Exported `NeonAuthServerApiError` from `@neondatabase/auth` and `@neondatabase/auth/next/server`.

packages/auth/src/server/client-factory.ts
Shridhad: Mint / cookie-validation warns now include `err` as suggested; updated stale SameSite=Lax comment to match `effectiveSameSite`.

packages/auth/src/server/proxy/request.ts
Shridhad: Split upstream HTTP logging: 4xx → `info`, 5xx → `warn` in proxy + server API client.

packages/auth/src/server/proxy/request.ts
Shridhad: Ensured trailing newline on `proxy/request.ts`. Left indent style as-is to avoid a huge whitespace-only diff (no repo Prettier config).

packages/auth/src/next/server/index.ts
Shridhad: Filled JSDoc gaps (`createNeonAuth` logger, `exchangeOAuthToken` log, `NeonAuthLogLevel` wording, exports on network-error types).

packages/auth/CHANGELOG.md
Shridhad: CHANGELOG lists all seven stable codes and points to `NEON_AUTH_NETWORK_ERROR_CODES` / `NeonAuthNetworkErrorCode`.

Shridhad (commented):
Took another pass at `85deb26`. Every Critical and Important item from the previous review is addressed, plus all the Suggestions I asked about. Tests are green (158 passing — 3 new: silent-overrides-custom-logger, DOMException AbortError, pre-resolved-log forwarding) and `tsc --noEmit` is clean.

### Verified fixes

| # | Item | Status |
|---|---|---|
| Critical | `auth.middleware()` honors `logger` / `logLevel` | ✅ `createNeonAuth` passes pre-resolved `log`; wrapper forwards all three; `processAuthMiddleware` prefers `config.log`; new test locks the wiring |
| Imp #2 | `kind: 'internal'` no longer leaks raw `Error.message` | ✅ `INTERNAL_PUBLIC_MESSAGE` constant + test updated |
| Imp #3 | Internal failures rethrow instead of POJO | ✅ `throw error instanceof Error ? error : new Error(String(error))` + CHANGELOG note |
| Imp #4 | `isAbortError` handles `DOMException` | ✅ name-only check on non-`Error` objects + new test |
| Imp #6 | `err: error` passed in log meta | ✅ all three sites (mint failure, cookie validation, server-API fetch) + JSDoc note on `NeonAuthLogger` |
| Imp #7 | `'silent'` + custom `logger` documented | ✅ `NeonAuthLoggingInput` JSDoc + new test |
| Imp #8 | No more per-request resolver allocation | ✅ `processor` prefers `config.log` |
| Imp #9 | Stale `SameSite=Lax` comment | ✅ updated to *"apply configured SameSite (default `strict`)"* |
| Imp #10 | JSDoc gaps | ✅ all sub-items addressed |
| Sug #1 | 4xx → `info`, 5xx → `warn` | ✅ both proxy/request and client-factory |
| Sug #2 | `ResolvedNeonAuthLogging = Required<NeonAuthLogger>` | ✅ + `NeonAuthActiveLogLevel` made private |
| Sug #3 | Typed `NeonAuthServerApiError` envelope | ✅ exported from main + next/server |
| Sug #4 | `NEON_AUTH_NETWORK_ERROR_CODES` semver-stable const | ✅ with JSDoc stability note |
| Sug #7 | Full code list in CHANGELOG | ✅ 7 codes enumerated, references to const + type |
| Sug #10 | Silent JSON parse swallow | ✅ logs at `debug` with diagnostic detail |

### Bonus improvements

- `classifyFetchFailure` is now publicly exported — lets adopters classify their own fetch failures with the same taxonomy.
- `NeonAuthConfig` / `NeonAuthMiddlewareConfig` refactored to type intersection with `NeonAuthLoggingInput`; cleaner shape, equivalent inference.
- CHANGELOG also documents the rethrow change and the 4xx/info, 5xx/warn split — addresses the severity-downgrade observability concern I'd flagged but skipped.

### One minor leftover (non-blocking)

Sug #9 — file-end newline is fixed, but `proxy/request.ts` and `middleware/processor.ts` still use tabs while `client-factory.ts` and `next/server/index.ts` use 2-space indent. Style nit; consider running prettier across the package in a follow-up.

LGTM for me — happy to merge from a review standpoint.

PR #139, opened 2026-05-12
docs(auth): fix authApiHandler JSDoc catch-all segment name

thekauer:
The JSDoc said to create the route at `app/api/auth/[...all]/route.ts`, but the handler reads `params.path.join('/')` — following the doc as written causes a runtime crash because `params.path` is undefined.

Align with `createNeonAuth` JSDoc, `llms.txt`, and the examples, which all use `[...path]`.

### DIFF
--- a/packages/auth/src/next/server/handler.ts
+++ b/packages/auth/src/next/server/handler.ts
@@ -17,11 +17,11 @@ type Params = { path: string[] };
  * @throws Error if `cookies.secret` is less than 32 characters
  *
  * @example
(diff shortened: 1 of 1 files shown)


thekauer:
Closing to clean up the stale PR queue. This is a low-value 2-line JSDoc-only change and the branch is 49 days stale with all CI red. Note: the underlying doc nit is still genuinely present on main - the authApiHandler JSDoc @example shows the catch-all route segment as [...all], but the handler destructures resolvedParams.path (type Params = { path: string[] }), so the example should read [...path]. Worth a tiny follow-up so it is not silently lost. Not deleting the branch.

PR #161, opened 2026-06-02
feat(auth): promote framework-agnostic server toolkit to public beta subpath (`@neondatabase/auth/server`)

Shridhad:
## Summary

Phase 1 of opening the Neon Auth server stack to community-built framework adapters. Promotes the framework-agnostic primitives the bundled `@neondatabase/auth/next/server` adapter is built on to a new public subpath: **`@neondatabase/auth/server`** (stability: **beta**).

This unblocks adapters for Hono, Remix, SolidStart, Express, Fastify, etc. without forking the package, and is the foundation for the in-flight TanStack Start adapter (#73) and the upcoming Hono adapter (Phase 2).

## Public surface (`@neondatabase/auth/server`)

28 exports, pinned by a snapshot fixture test (`packages/auth/src/server/public-surface.test.ts`):

| Category | Exports |
|---|---|
| Factory | `createAuthServer` |
| Proxy | `handleAuthProxyRequest`, `handleAuthRequest`, `handleAuthResponse` |
| Middleware | `processAuthMiddleware`, `shouldProtectRoute`, `checkSessionRequired`, `DEFAULT_AUTH_SKIP_ROUTES` |
| Config | `validateCookieConfig`, `resolveNeonAuthLogging`, `classifyFetchFailure` |
| Session | `parseSessionData`, `validateSessionData` |
| Cookies | `parseSetCookies`, `serializeSetCookie`, `parseCookieValue`, `extractNeonAuthCookies` |
| Errors | `AuthError`, `AuthApiError`, `isAuthError`, `isAuthApiError` |
| Constants | `NEON_AUTH_SERVER_PROXY_HEADER`, `NEON_AUTH_HEADER_MIDDLEWARE_NAME`, `NEON_AUTH_COOKIE_PREFIX`, `NEON_AUTH_SESSION_COOKIE_NAME`, `NEON_AUTH_SESSION_DATA_COOKIE_NAME`, `NEON_AUTH_SESSION_CHALLENGE_COOKIE_NAME`, `NEON_AUTH_NETWORK_ERROR_CODES` |
| Types | `RequestContext`, `RequestContextFactory`, `CookieOptions`, `NeonAuthConfig`, `NeonAuthMiddlewareConfig`, `AuthMiddlewareConfig`, `AuthProxyConfig`, `MiddlewareResult`, `SessionData`, `SessionCookieConfig`, `SessionCookieSameSite`, and others |

## Internal-only changes

- **Rename** `createAuthServerInternal` -> `createAuthServer`. Symbol was never exported; safe direct rename (no alias).
- **Promote** the Next.js `SKIP_ROUTES` constant to the shared `DEFAULT_AUTH_SKIP_ROUTES`. Next adapter now consumes the shared constant. Skip list is byte-identical and pinned by `route-protection.test.ts`.
- **Widen** `AuthMiddlewareConfig.skipRoutes` (and `shouldProtectRoute` / `checkSessionRequired`) from `string[]` to `readonly string[]`. Non-breaking — accepts both mutable arrays and the new `as const` tuple.
- **Drop** `@internal` from `parseSessionData` (was tripping `stripInternal` once it became a public re-export; now correctly `@public` with a substantive contract).
- **JSDoc** full contract added to every `RequestContext` interface member.

## Docs

- New **`packages/auth/BUILDING-AN-ADAPTER.md`** — 4-step adapter author guide with Hono code examples.
- README gains a "Server toolkit (for framework adapter authors)" section with the `@beta` notice.
- CHANGELOG entry under `[Unreleased]`.

## Stability commitment

`@neondatabase/auth/server` ships as **beta**. While in beta:
- We will avoid breaking changes whenever possible.
- Any breaking change ships in a minor version (`0.x.0`) with a CHANGELOG entry and one-step migration.
- When stable we'll move it to stable in a minor release.

## Version

`@neondatabase/auth`: `0.4.1-beta` -> **`0.5.0-beta`**

## Test plan

- [x] Repo-wide typecheck: all 13 packages clean
- [x] Repo-wide build: green; `@neondatabase/neon-js` workspace dep resolves to `^0.5.0-beta`
- [x] Repo-wide tests: 208/208 passing (auth 163, auth-ui 28, neon-js 17)
- [x] `dist/next/server/index.mjs` public exports byte-identical to pre-change baseline
- [x] No `@/` alias leaks in `dist/`
- [x] Node ESM smoke-import of built `dist/server/index.mjs` resolves all 28 exports
- [x] Public surface snapshot test locks in the export list
- [x] `DEFAULT_AUTH_SKIP_ROUTES` pin test asserts byte-identity with legacy `SKIP_ROUTES`
- [ ] CI: lint, build, typecheck, unit-tests, E2E (will run on PR open)

## Next phases (not in this PR)

- **Phase 2**: Hono adapter (`@neondatabase/auth/hono`) + `examples/hono-neon-auth`, built entirely on top of this toolkit
- Coordinate with TanStack PR #73 to rebase onto this toolkit


packages/auth/BUILDING-AN-ADAPTER.md
```diff
@@ -0,0 +1,226 @@
...
+import { extractNeonAuthCookies } from '@neondatabase/auth/server';
+import { getContext } from 'hono/context-storage'; // example: Hono
+
+function createHonoRequestContext(): RequestContext {
+  const c = getContext();
+  return {
+    getCookies: () => extractNeonAuthCookies(c.req.header('cookie') ?? ''),
+    setCookie: (name, value, options) => {
+      // Map toolkit options into your framework's cookie API
+      c.header('Set-Cookie', serializeCookie(name, value, options), {
```
Shridhad: `serializeCookie` isn't exported from `@neondatabase/auth/server` — the helper you want is `serializeSetCookie`, which takes a single object rather than positional `(name, value, options)` args:

```typescript
import { serializeSetCookie } from '@neondatabase/auth/server';

setCookie: (name, value, options) => {
  c.header('Set-Cookie', serializeSetCookie({ name, value, ...options }), { append: true });
},
```

packages/auth/BUILDING-AN-ADAPTER.md
```diff
@@ -0,0 +1,226 @@
...
+`/api/auth/*` path, extracting `paths` from your framework's router:
+
+```typescript
+// handler.ts
+import { handleAuthProxyRequest } from '@neondatabase/auth/server';
+
+export function authHandler(config: ProxyConfig) {
+  return async (c: Context) => {
+    const path = c.req.path.split('/api/auth/')[1]?.split('/') ?? [];
+    return handleAuthProxyRequest(c.req.raw, path, config);
```
Shridhad: Two issues here:

1. `handleAuthProxyRequest` takes a single `AuthProxyConfig` object, not positional arguments — `(c.req.raw, path, config)` won't compile. The description on line 120 (`"pure function from (Request, paths, config)"`) is also wrong.
2. `path` should be a `string` (e.g. `'sign-in/email'`), not `string[]`.

Correct form:
```typescript
const path = c.req.path.replace(/^\/api\/auth\//, '');
return handleAuthProxyRequest({
  request: c.req.raw,
  path,
  baseUrl: config.baseUrl,
  cookieSecret: config.cookieSecret,
  log,
});
```

Ref: `src/server/proxy/handler.ts` line 41 for the full `AuthProxyConfig` shape.

packages/auth/BUILDING-AN-ADAPTER.md
```diff
@@ -0,0 +1,226 @@
...
+  return async (c: Context, next: Next) => {
+    const url = new URL(c.req.url);
+    const result = await processAuthMiddleware({
+      request: c.req.raw,
+      pathname: url.pathname,
+      skipRoutes: DEFAULT_AUTH_SKIP_ROUTES,
+      ...config,
+    });
+
+    switch (result.type) {
```
Shridhad: Several bugs in this switch block that will silently break any adapter built from this example:

- `result.type` → should be `result.action` (the discriminant on `MiddlewareResult`). At runtime `result.type` is always `undefined`, so every case is unreachable.
- `result.cookiesToSet` → `result.cookies` (`string[] | undefined`)
- `case 'redirect'` doesn't exist — the two variants are `'redirect_oauth'` and `'redirect_login'`
- `result.url` → `result.redirectUrl` (`URL`, not `string`)

Critically, `'redirect_oauth'` is missing entirely. Without it, post-OAuth `?code=` callbacks never complete and users silently loop back to login.

Correct pattern (mirrors `src/next/server/middleware.ts`):
```typescript
switch (result.action) {
  case 'allow':
    await next();
    for (const cookie of result.cookies ?? []) c.header('Set-Cookie', cookie, { append: true });
    return;
  case 'redirect_oauth':
    for (const cookie of result.cookies) c.header('Set-Cookie', cookie, { append: true });
    return c.redirect(result.redirectUrl.toString(), 302);
  case 'redirect_login':
    return c.redirect(result.redirectUrl.toString(), 302);
}
```

packages/auth/package.json
```diff
@@ -78,6 +78,10 @@
       "types": "./dist/next/server/index.d.mts",
       "default": "./dist/next/server/index.mjs"
     },
+    "./server": {
```
Shridhad, line 81: The monorepo convention (CLAUDE.md) is that `@neondatabase/neon-js` re-exports every `@neondatabase/auth` subpath under `@neondatabase/neon-js/auth/*`. This new `./server` subpath isn't wired up there — adapter authors using the full SDK would need to add `@neondatabase/auth` as a separate dependency to reach the server toolkit.

Needs three additions:
- `packages/neon-js/src/auth/server/index.ts` re-exporting from `@neondatabase/auth/server`
- Entry in `packages/neon-js/package.json` exports: `"./auth/server": { ... }`
- Entry in `packages/neon-js/tsdown.config.ts`

packages/auth/BUILDING-AN-ADAPTER.md
Shridhad: Fixed in fa7841bf1605e0eb38e99006c9134e712453c449. Step 1 cookie example now uses `serializeSetCookie({ name, value, ...options })` (the actual exported helper, takes one `ParsedCookie` object).

packages/auth/BUILDING-AN-ADAPTER.md
Shridhad: Fixed in fa7841bf1605e0eb38e99006c9134e712453c449. Step 3 example now calls `handleAuthProxyRequest({ request, path, baseUrl, cookieSecret, ... })` with `path` as a slash-joined `string` (`c.req.path.replace(/^\/api\/auth\//, '')`). Prose claim on line 120 corrected too.

packages/auth/BUILDING-AN-ADAPTER.md
Shridhad: Fixed in fa7841bf1605e0eb38e99006c9134e712453c449. Step 4 middleware rewritten to mirror `src/next/server/middleware.ts` exactly: discriminant `action`, all three variants (`'allow'`, `'redirect_oauth'`, `'redirect_login'`), correct field names (`cookies` / `redirectUrl`), and `result.headers` forwarding on `'allow'` for the `x-neon-auth-middleware` signal. Added a variants table at the top so the OAuth-callback path is hard to miss.

packages/auth/package.json
Shridhad: Fixed in fa7841bf1605e0eb38e99006c9134e712453c449. Wired `./auth/server` into `@neondatabase/neon-js` per the convention:\n- `packages/neon-js/src/auth/server/index.ts` re-exports `@neondatabase/auth/server`\n- `packages/neon-js/package.json` exports gains `./auth/server`\n- `packages/neon-js/tsdown.config.ts` gains the new entry\n- Version bump `0.6.1-beta` → `0.7.0-beta` + CHANGELOG entry\n- CLAUDE.md updated with the new export line and a 5-step "whenever you add a new auth subpath" checklist so the convention is harder to miss in future\n\nVerified: `import * as t from '@neondatabase/neon-js/auth/server'` resolves and surfaces all 28 toolkit symbols (Node ESM smoke test).

thekauer:
## Review: server toolkit public-beta surface

Reviewed with a harsh self-review pass plus a 3-model ace pass (gpt-5.5 / gpt-5.4 / claude). Net: no blocking runtime bugs (build green, 134 server tests pass, the 28 runtime exports resolve from the built `dist/server/index.mjs`, no `@/` alias leaks). This is a request-changes on the author-facing docs and the public API-surface guarantees, not on shipped runtime behavior.

### Must fix before widening the beta

**1. `BUILDING-AN-ADAPTER.md` has two non-compiling examples** (all reviewers and all 3 ace models agree). Since enabling third-party adapter authors is the whole point of this PR, the guide must compile.
- Middleware example (around line 160) switches on `result.type` with cases `'allow'` / `'redirect'` and reads `result.cookiesToSet` / `result.url`. The real `MiddlewareResult` discriminates on `action` with `'allow' | 'redirect_oauth' | 'redirect_login'` and exposes `cookies` / `redirectUrl`. None of `type`, `'redirect'`, `cookiesToSet`, `url` exist. Mirror `next/server/middleware.ts`, which switches on `result.action`. As written, copied code never matches a case and silently never redirects unauthenticated users (a broken auth gate).
- Handler example (around line 120) calls `handleAuthProxyRequest(c.req.raw, path, config)` with 3 positional args and `path` as an array. Real signature is `handleAuthProxyRequest(config: AuthProxyConfig)` (single object, `path: string`). See `next/server/handler.ts`, which does `path = resolvedParams.path.join('/')` then passes one config object.

**2. `createAuthServer` does no `cookieSecret` validation (HIGH).** The now-public factory signs and validates `session_data` cookies with this secret but never checks length or emptiness. The Next path guards via `validateCookieConfig`; direct toolkit consumers get no guard, so a weak or empty secret yields forgeable session cookies. Add the same validation inside `createAuthServer`.

**3. Public type surface is unpinned and undocumented.** `public-surface.test.ts` snapshots only the 28 runtime `Object.keys`, so the ~15 type-only exports can be renamed or dropped silently while the test stays green. Add a `*.test-d.ts` or packaged `.d.mts` smoke check. Also `NeonAuthServerConfig` is now public (required to call `createAuthServer`) but has zero JSDoc and is missing from the CHANGELOG export list.

### Pre-existing, but now public (document or harden)
- `shouldProtectRoute` (route-protection.ts) uses bare `startsWith`, so a skip prefix `/auth/sign-in` also skips `/auth/sign-internal`, and `/api/auth` skips `/api/authz`. It is now the exported default skip-matcher for adapters. Consider segment-aware matching.
- Proxy leaf methods forward only `(path, method, args)` and discard `fetchOptions` (`_fetchOptions`), while the type claims Better Auth compatibility where `method(data, fetchOptions)` is valid. Either honor it or document the limitation.
- `parseSessionData` (now public) writes to `console.error` unconditionally, bypassing the package's configurable `logLevel: 'silent'` logger.

### Nits
- `AuthError` / `AuthApiError` and their guards are re-exported straight from `@supabase/auth-js` into the vendor-neutral public surface; this couples the contract to a pinned Supabase version and risks `instanceof`-across-realms failures. Consider neon-owned wrappers.
- `route-protection.test.ts` asserts `Object.isFrozen(...)` is `false` (a vacuous assertion).
- `package.json` `files` whitelist omits `BUILDING-AN-ADAPTER.md`, so the guide is not published to npm.
- `operations.ts` uses `json as any` in now-public code.

Build, typecheck, and lint are all clean and the runtime code is sound. Items 1-3 are the gates before promoting the beta widely.

Shridhad:
Thanks for the thorough review @thekauer — all 11 items addressed in 0657f535d0a01c98a5c39c4ce71faf91cd09cef9. Scorecard:

### Must-fix
- **#1 doc compile (middleware)** — already fixed in `fa7841b` (same bug @shridhar-deshmukh flagged inline). Verified the example now mirrors `next/server/middleware.ts` exactly (switches on `result.action`, three variants, correct field names).
- **#2 doc compile (handler)** — already fixed in `fa7841b`. Now uses `handleAuthProxyRequest({ ... })` with `path: string`.
- **#3 `createAuthServer` cookie-secret validation (HIGH)** — fixed. The factory now calls `validateCookieConfig({ secret: cookieSecret, sessionDataTtl })` so weak (<32 char) / empty secrets and non-positive TTLs throw at construction time. +4 tests in `client-factory.test.ts`.
- **#3.2 `NeonAuthServerConfig` JSDoc + CHANGELOG export list** — fixed. Each field gets intent + defaults + examples + validation behavior; CHANGELOG export list amended.
- **#3.3 type-only export pinning** — fixed via new `public-surface-types.test.ts`. Imports all 24 type-only exports by name (rename/removal caught by `tsc --noEmit`) and pins the shape of the five highest-value contract types (`NeonAuthServerConfig`, `RequestContext`, `MiddlewareResult`, `AuthProxyConfig`, `NeonAuthServerApiError`) with `expectTypeOf`. Companion to `public-surface.test.ts`.

### Pre-existing-but-now-public
- **#4 `shouldProtectRoute` prefix bleed** — fixed. Switched from `pathname.startsWith(route)` to segment-aware matching (`pathname === route || pathname.startsWith(route + '/')`), with trailing-slash normalization. +18 regression tests covering the exact bug examples (`/auth/sign-internal`, `/api/authz`, `/auth/callback-handler`, `/auth/sign-uphill`, `/auth/forgot-password-recovery`).
- **#5 `fetchOptions` discarded by server proxy** — documented as an intentional limitation under a new "Known limitations" section in `BUILDING-AN-ADAPTER.md`. Rationale: client-side hooks (`onSuccess`/`onError`/`throw`) don't translate server-side; forwarding arbitrary `headers` would bypass the toolkit's cookie sanitization. `signal`/`query` flagged as candidates for a future iteration.
- **#6 `parseSessionData` bypasses `logLevel: 'silent'`** — fixed across the whole session cache path. Threaded optional `log: ResolvedNeonAuthLogging` through `parseSessionData`, `getSessionDataFromCookie`, `fetchSessionWithCookie`, `mintSessionDataFromResponse`, `mintSessionDataFromToken`, `trySessionCache`, `handleAuthResponse`. `createAuthServer` + `handleAuthProxyRequest` plumb it in automatically. +2 tests assert the logger is honored AND no `console.error` fallback leaks when one is provided. Back-compat preserved (optional param; absent = legacy `console.*`).

### Nits
- **#7 Supabase `AuthError` coupling** — fixed by dropping `AuthError`/`AuthApiError`/`isAuthError`/`isAuthApiError` from `@neondatabase/auth/server`. Toolkit is now vendor-neutral; adapter authors narrow on `NeonAuthServerApiError.code` (`NETWORK_*` | `INTERNAL_ERROR` | `string & {}`) instead of `instanceof AuthError`. Symbols remain on `@neondatabase/auth`, `@neondatabase/auth/next`, `@neondatabase/auth/next/server`, `@neondatabase/auth/vanilla/adapters` for SupabaseAuthAdapter consumers. Runtime surface: 28 → 24 symbols. Snapshot updated. `BUILDING-AN-ADAPTER.md` gains an error-narrowing example. I went with "drop entirely" rather than "wrap" since the toolkit code doesn't actually emit these errors itself (errors come through `normalizeBetterAuthError`); neon-owned wrappers can land if a concrete need emerges.
- **#8 vacuous `Object.isFrozen` assertion** — removed during the #4 rewrite. The `as const` tuple's readonly-ness is enforced by the type system; the runtime check was asserting the unintuitive runtime behavior, not contract coverage.
- **#9 `BUILDING-AN-ADAPTER.md` missing from npm tarball** — fixed. Added to `package.json` `files`. Verified via `npm pack --dry-run` (12.7kB ships, listed alongside README.md).
- **#10 `json as any`** — fixed. Narrowed to `RawSessionEnvelope` shape that documents exactly which fields `parseSessionData` inspects (open-ended via index signatures for forward compatibility on other Better Auth fields).

### Verification harness (from a clean working tree)
- `pnpm typecheck` — green
- `pnpm lint` — green
- `pnpm --filter @neondatabase/auth test:ci` — **203/203** (was 153; +50 new tests)
- `pnpm test:ci` (full repo) — green (auth 203, postgrest-js 28, neon-js 17 with typecheck)
- `pnpm --filter @neondatabase/auth build` — green (`dist/server/index.mjs` = 1.3 kB)
- `pnpm --filter @neondatabase/neon-js build` — green
- `npm pack --dry-run` — `BUILDING-AN-ADAPTER.md` (12.7kB) confirmed in tarball
- Built-dist smoke: `import('./dist/server/index.mjs')` → 24 symbols, `AuthError` absent, `createAuthServer` callable
- End-to-end via umbrella: `shouldProtectRoute('/auth/sign-internal', DEFAULT_AUTH_SKIP_ROUTES) === true` (the exact prefix-bleed case you flagged, now correctly protected through `@neondatabase/neon-js/auth/server`)

PTAL when you have a moment 🙏

Shridhad:
Status update — PR rebased and CI is running fresh:

- Rebased onto `main` (resolved 2 trivial version-bump conflicts: kept `@neondatabase/auth@0.5.0-beta` and `@neondatabase/neon-js@0.7.0-beta` over `main`'s 0.4.2-beta / 0.6.2-beta patch bumps)
- Force-pushed `0657f53 → 486db0a` with `--force-with-lease`
- Local verification post-rebase: typecheck/lint/build/tests all green; auth 203/203; end-to-end smoke through `@neondatabase/neon-js/auth/server` still surfaces 24 symbols and `shouldProtectRoute('/auth/sign-internal', DEFAULT_AUTH_SKIP_ROUTES) === true` (Andras's prefix-bleed case correctly protected)
- GH Actions now running on `486db0a` (didn't trigger on yesterday's push — likely a transient hiccup, resolved by the rebase)

@thekauer @andrelandgraf — ready for re-review when you have a moment 🙏. The 11-item scorecard for Andras's earlier review is [here](https://github.com/neondatabase/neon-js/pull/161#issuecomment-4650829162); all items addressed in `0657f53` (now `486db0a` after rebase).

thekauer (commented):
Round of review fixes from authoring 5 real adapters (Hono, Remix, SolidStart, Express, Fastify) against this branch - each completes a full sign-in -> protected-route -> 401 round-trip, so these are grounded in actual adapter authoring rather than a read-through. All code suggestions were validated locally (`tsc --noEmit` clean, `vitest run src/server` 180 passing).

2 must-fix (Secure forced on the forwarding paths to match minting; exact cookie-name matching in minting), the rest are docs/DX + a small correctness nit. Nice work on the toolkit - the adapter surface came together cleanly. Inline comments below.

packages/auth/src/server/client-factory.ts
```diff
@@ -163,7 +254,8 @@ export function createAuthServerInternal(
             sessionDataTtl,
             domain,
             sameSite,
-          }
+          },
+          log
```
thekauer, line 276: **FIX 1 (must-fix, security): force `Secure` on this forwarding path.**

Just above this (the `cookieOptions` spread around the upstream-cookie loop), the forwarded first-party cookies inherit `secure` from the upstream value and don't force it, while `session/minting.ts` hardcodes `secure: true` and the docs promise "Secure is always applied." When `effectiveSameSite === 'none'` a missing `Secure` makes the browser drop the cookie entirely. Add `secure: true` to the spread:

```ts
const cookieOptions = {
  ...cookie,
  domain: domain,
  partitioned: undefined,
  sameSite: effectiveSameSite,
  // Always force Secure, matching the minting path (session/minting.ts) and the
  // documented "Secure is always applied" contract. Required for SameSite=None.
  secure: true,
};
```

(Verified: `tsc --noEmit` clean + `vitest run src/server` 180 passing on this branch.)

packages/auth/src/server/proxy/response.ts
```diff
@@ -20,12 +21,13 @@ const RESPONSE_HEADERS_ALLOWLIST = ['content-type', 'content-length', 'content-e
 export const handleAuthResponse = async (
```
thekauer, line 24: **FIX 1 (must-fix, security): force `Secure` in `prepareResponseHeaders`.**

In the `set-cookie` branch of `prepareResponseHeaders` below, each forwarded cookie gets `partitioned = undefined` and `sameSite = effectiveSameSite` but keeps whatever `secure` the upstream sent. That contradicts the minting path (always `secure: true`) and the "Secure is always applied" docs, and silently breaks `SameSite=None`. Add one line in that loop, right after the `sameSite` assignment:

```ts
parsedCookie.partitioned = undefined;
parsedCookie.sameSite = effectiveSameSite;
// Always force Secure on forwarded cookies, matching the minting path and the
// "Secure is always applied" contract. Required for SameSite=None, harmless otherwise.
parsedCookie.secure = true;
if (domain) {
  parsedCookie.domain = domain;
}
```

(I added two tests for this in `response.test.ts`: a forwarded cookie with no upstream `Secure`, and a `SameSite=None` case; both pass.)

packages/auth/src/server/session/minting.ts
```diff
@@ -74,7 +84,8 @@ async function mintSessionDataCookie(
 export async function mintSessionDataFromResponse(
   responseHeaders: Headers,
   baseUrl: string,
-  cookieConfig: SessionCookieConfig
+  cookieConfig: SessionCookieConfig,
+  log?: ResolvedNeonAuthLogging
 ): Promise<string | null> {
   // Check if upstream set a session_token cookie
   const setCookieHeaders = responseHeaders.getSetCookie();
```
thekauer, line 100: **FIX 2 (must-fix, correctness): exact cookie-name match instead of substring scans.**

Just below, the session-token detection does `setCookieHeaders.find(c => c.includes('session_token'))` and the sign-out check does `.toLowerCase().includes('max-age=0')` over the raw Set-Cookie text. Both false-match: an unrelated cookie named e.g. `analytics_session_token_ref` matches the first, and a token whose *value* contains the literal `max-age=0` (with a real `Max-Age=3600`) is wrongly treated as a sign-out. Parse with the existing `parseSetCookies` helper and match the exact `NEON_AUTH_SESSION_COOKIE_NAME`, reading sign-out from the parsed `maxAge`/`expires`:

```ts
import {
  NEON_AUTH_SESSION_COOKIE_NAME,
  NEON_AUTH_SESSION_DATA_COOKIE_NAME,
} from '../constants';
import { parseSetCookies, serializeSetCookie } from '../utils/cookies';

// ...inside mintSessionDataFromResponse:
const setCookieHeaders = responseHeaders.getSetCookie();
let sessionTokenCookie: string | undefined;
let sessionTokenIsDeletion = false;
for (const cookieHeader of setCookieHeaders) {
  const parsed = parseSetCookies(cookieHeader).find(
    cookie => cookie.name === NEON_AUTH_SESSION_COOKIE_NAME
  );
  if (parsed) {
    sessionTokenCookie = cookieHeader;
    sessionTokenIsDeletion =
      parsed.maxAge === 0 ||
      (parsed.expires !== undefined && parsed.expires.getTime() <= Date.now());
    break;
  }
}
if (!sessionTokenCookie) return null;
if (sessionTokenIsDeletion) { /* return deletion cookie */ }
```

Note: `parseSetCookies` doesn't currently surface `expires`, so add `expires: parsedCookie.expires` in `utils/cookies.ts` (the `ParsedCookie` field already exists) to make the past-`Expires` check live. (Added 3 false-match/deletion tests in `minting.test.ts`; all pass.)

packages/auth/src/server/proxy/handler.ts
```diff
@@ -61,10 +66,15 @@ export async function handleAuthProxyRequest(config: AuthProxyConfig): Promise<R
 
 	// Fallback: Call upstream API
 	const response = await handleAuthRequest(baseUrl, request, path, log);
```
thekauer, line 80: **FIX 3 (DX): `AuthProxyConfig.log` is effectively `Required<NeonAuthLogger>`, so a public `NeonAuthLogger` passthrough fails TS2322.**

The public `NeonAuthLogger` is `Partial<...>` but `AuthProxyConfig.log` is `ResolvedNeonAuthLogging` (= `Required<NeonAuthLogger>`). An adapter forwarding a `log?: NeonAuthLogger` straight into `handleAuthProxyRequest` (exactly the Step 3 doc pattern) doesn't compile. Minimal fix: widen the field and normalize once internally, preserving the existing `undefined`/pre-resolved passthrough so behavior is unchanged for `createNeonAuth`:

```ts
// AuthProxyConfig
log?: ResolvedNeonAuthLogging | NeonAuthLogger;

// in handleAuthProxyRequest, before use:
const log = resolveLog(config.log);

function resolveLog(
  log: ResolvedNeonAuthLogging | NeonAuthLogger | undefined
): ResolvedNeonAuthLogging | undefined {
  if (!log) return undefined; // unchanged: downstream silence
  if (log.error && log.warn && log.info && log.debug) {
    return log as ResolvedNeonAuthLogging; // pre-resolved: keep level gating
  }
  return resolveNeonAuthLogging({ logger: log });
}
```

(Confirmed the TS2322 repro before/after; full server suite stays green, including the handler tests that assert `undefined` is forwarded when no logger is set. Added a `toMatchTypeOf` pin in `public-surface-types.test.ts`.)

packages/auth/src/server/proxy/response.ts
```diff
@@ -1,6 +1,7 @@
 import { mintSessionDataFromResponse } from '../session/minting';
 import { parseSetCookies, serializeSetCookie } from '@/server/utils/cookies';
 import type { SessionCookieConfig } from '../config';
+import type { ResolvedNeonAuthLogging } from '../logger';
 
 // Allowlist of response headers that we want to proxy to the client from Neon Auth.
 const RESPONSE_HEADERS_ALLOWLIST = ['content-type', 'content-length', 'content-encoding', 'transfer-encoding',
```
thekauer: **FIX 4 (nit, correctness): drop hop-by-hop / framing headers from the allowlist.**

`connection`, `transfer-encoding`, and `content-length` shouldn't be copied onto the new `Response` - the runtime sets framing itself, and a forwarded stale `content-length` corrupts a re-encoded body. Keep content-type/content-encoding/set-cookie/set-auth-*/request-id:

```ts
// Allowlist of response headers that we want to proxy to the client from Neon Auth.
// Hop-by-hop / framing headers (connection, transfer-encoding, content-length) are
// intentionally excluded: the runtime sets framing, and a stale content-length corrupts
// a re-encoded body.
const RESPONSE_HEADERS_ALLOWLIST = ['content-type', 'content-encoding', 'date',
   'set-cookie', 'set-auth-jwt', 'set-auth-token', 'x-neon-ret-request-id'];
```

packages/auth/BUILDING-AN-ADAPTER.md
```diff
@@ -0,0 +1,322 @@
...
+    sameSite: config.cookies.sameSite,
+    log,
+  });
+
+  // Attach the route handler + middleware so apps get one cohesive surface.
+  Object.assign(server, {
+    handler: () => authHandler({ baseUrl: config.baseUrl, log, ...config.cookies }),
+    middleware: (opts: { loginUrl: string }) =>
+      authMiddleware({ baseUrl: config.baseUrl, log, ...config.cookies, ...opts }),
+  });
```
thekauer, line 129: **FIX 5 (docs): this getting-started snippet doesn't compile.**

`Object.assign` spreads `...config.cookies`, which yields `{ secret, ... }`, but `authHandler` / `authMiddleware` (Steps 3 & 4) read FLAT fields (`config.cookieSecret`, `config.domain`, `config.sameSite`). So `cookieSecret` ends up `undefined` and the spread shape never matches the `ProxyConfig` the handlers expect. Map `config.cookies` explicitly into one flat `proxyConfig` and reuse it (this also matches the real flat `NeonAuthServerConfig` = `{ baseUrl, cookieSecret, context, sessionDataTtl?, domain?, sameSite?, log? }`):

```suggestion
  // `createAuthServer` takes a FLAT config: the nested `cookies` block from
  // your adapter's public `NeonAuthConfig` is mapped to top-level fields.
  const server = createAuthServer({
    baseUrl: config.baseUrl,
    context: createHonoRequestContext,
    cookieSecret: config.cookies.secret,
    sessionDataTtl: config.cookies.sessionDataTtl,
    domain: config.cookies.domain,
    sameSite: config.cookies.sameSite,
    log,
  });

  // Attach the route handler + middleware so apps get one cohesive surface.
  // `authHandler` / `authMiddleware` (Steps 3 & 4) also read FLAT fields
  // (`cookieSecret`, `domain`, `sameSite`), so map `config.cookies` explicitly
  // rather than spreading it (a spread would yield `secret`, not `cookieSecret`).
  const proxyConfig = {
    baseUrl: config.baseUrl,
    cookieSecret: config.cookies.secret,
    sessionDataTtl: config.cookies.sessionDataTtl,
    domain: config.cookies.domain,
    sameSite: config.cookies.sameSite,
    log,
  };

  Object.assign(server, {
    handler: () => authHandler(proxyConfig),
    middleware: (opts: { loginUrl: string }) =>
      authMiddleware({ ...proxyConfig, ...opts }),
  });
```

packages/auth/BUILDING-AN-ADAPTER.md
```diff
@@ -0,0 +1,322 @@
...
+          for (const cookie of result.cookies) c.header('Set-Cookie', cookie, { append: true });
+        }
+        return c.redirect(result.redirectUrl.toString(), 302);
+    }
+  };
+}
+```
+
+`DEFAULT_AUTH_SKIP_ROUTES` matches what the Next.js adapter uses; you can
+compose it with your own public routes (`[...DEFAULT_AUTH_SKIP_ROUTES, '/healthz']`).
```
thekauer, line 234: **FIX 6 (docs/DX): add a "Node-style (non-Fetch) frameworks" section.**

Express and Fastify are listed as targets up top but get no guidance, and they hit several real gotchas I ran into building the adapters. Suggest inserting a new `## Node-style (non-Fetch) frameworks (Express, Fastify)` section here (before "What the toolkit gives you for free") covering, with snippets:

1. **Build a Fetch `Request` from a Node req** - absolute URL from `x-forwarded-proto` + `host`, copy headers, pass raw body, and set `duplex: 'half'` when there's a body (else Node's `Request` throws).
2. **Write the `Response` back, preserving multiple `Set-Cookie`** - CRITICAL: use `response.headers.getSetCookie()` (array), NOT `.get('set-cookie')`, which comma-merges and corrupts any cookie with `Expires=Wed, 01 Jan ...`. Write via `res.setHeader('set-cookie', cookies)` / `reply.raw.setHeader`.
3. **Raw body parsing** - Express: `express.raw({ type: '*/*' })` on the auth route; Fastify: `removeAllContentTypeParsers()` + `addContentTypeParser('*', { parseAs: 'buffer' }, ...)`.
4. **Express 5 wildcard** - needs a NAMED wildcard `app.all('/api/auth/*splat', ...)`; bare `/api/auth/*` throws at registration (path-to-regexp v8).
5. **Bind the request via `AsyncLocalStorage`** - the `context` factory runs lazily with no event access, so non-ambient frameworks must stash the per-request Fetch `Request` in an `AsyncLocalStorage` and have `RequestContext` read it.
6. **ESM-only** - `require()` works on Node `>= 22.12` via interop; otherwise use `await import('@neondatabase/auth/server')`.

I have the full section written with code blocks for each point if useful - happy to drop it in a follow-up or paste it here.

packages/auth/src/server/public-surface-types.test.ts
```diff
@@ -0,0 +1,218 @@
...
+      'NeonAuthNetworkErrorCode',
+      'ClassifiedFetchFailure',
+      'NeonAuthServer',
+      'NeonAuthServerApiError',
+      'SessionData',
+      'SessionDataCookie',
+      'RequireSessionData',
+      'ParsedCookie',
+    ];
+    expect(exportedTypeNames).toHaveLength(24);
```
thekauer: This assertion is a runtime tautology: it checks that a hardcoded 24-element array literal has 24 elements. It never references the imported types, so renaming or removing one of the 24 type exports does not make it fail.

I verified this empirically on this branch. I renamed the `ParsedCookie` re-export in `src/server/index.ts` so `import type { ParsedCookie } from './index'` dangles, then ran the three mechanisms against just this file:

| Mechanism | Caught the broken export? |
| --- | --- |
| `vitest run` (the `test:ci` script) | No - still PASSES, 10/10 tests green |
| `tsc --noEmit` (the `test:types` script) | Yes - FAILS with `public-surface-types.test.ts(69,3): error TS2305: Module '"./index"' has no exported member 'ParsedCookie'.` |
| `vitest run --typecheck` | Yes, but only once this file is in the `typecheck.include` glob (it FAILS with `TypeCheckError: Module '"./index"' has no exported member 'ParsedCookie'`). With vitest's default glob (`*.test-d.ts`) it does NOT cover this `*.test.ts` file, so the default `--typecheck` passed too. |

Under `test:ci` (`vitest run`, no `--typecheck`, and `vitest.config.ts` has no `typecheck` block) the `import type {...}` and `expectTypeOf<...>()` are erased to no-ops. So the only thing actually protecting this surface today is `tsc --noEmit` (the `test:types` script) running over this file.

Worth noting the distinction vs the sibling runtime snapshot: for runtime VALUE exports (`createAuthServer`, etc.) the `typeof X === 'function'` check is valid and sufficient. But the 24 here are TYPES (erased at runtime), so there is no runtime shortcut - they need a real typecheck.

Suggested fixes:
1. Drop the `toHaveLength(24)` tautology (suggestion below).
2. Make the type checks actually run: enable vitest `typecheck` (add a `typecheck` block to the config and include this file, so `expectTypeOf` executes), or make sure CI runs `pnpm test:types` over this file. As-is, `test:ci` alone gives no protection.
3. For true public-surface protection, import from the built specifier `@neondatabase/auth/server` (and `@neondatabase/neon-js/auth/server`) instead of `./index`, so the package `exports` map / dist `.d.ts` is verified too - ties into the earlier exports-resolution note.

```suggestion
```

packages/auth/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@neondatabase/auth",
-  "version": "0.4.2-beta",
+  "version": "0.5.0-beta",
```
thekauer: This PR manually bumps `@neondatabase/auth` 0.4.2-beta -> 0.5.0-beta (and `@neondatabase/neon-js` 0.6.2-beta -> 0.7.0-beta). Isn't version bumping owned by the Prepare Release workflow? I think let's revert this and let the release workflow handle it

Shridhad:
@thekauer thanks for the second round — five real adapters is *exactly* the validation the docs needed. All six items are addressed in `4c5425c`. Verification numbers up front, scorecard below.

**Verification**
- `pnpm typecheck` (workspace): clean
- `pnpm exec eslint packages/auth`: clean
- `vitest run` in `packages/auth`: **216 passing** (+8 vs. baseline 208 — covers all new behaviors)
- `vitest run src/server`: **187 passing** (matches your local 180 baseline + 7 new)
- `tsdown` build: clean; smoke import of `@neondatabase/auth/server` and `@neondatabase/neon-js/auth/server` both resolve all 24 runtime exports.

### FIX scorecard

| # | What | Where | Tests |
|---|------|-------|-------|
| **1 (security)** | Force `Secure: true` on forwarded cookies — proxy + factory paths, matching `session/minting.ts` and the documented contract | [`proxy/response.ts`](https://github.com/neondatabase/neon-js/pull/161/files#diff-3f0b6fbe9b3e5e8c2f31fb38bbf1eb1f) line 60–66, [`client-factory.ts`](https://github.com/neondatabase/neon-js/pull/161/files#diff-12b5d2a3a6cf6c1c7c1fa6c1a3a7e7a7) line 237–248 | **+4 tests**: 2 in [`response.test.ts`](https://github.com/neondatabase/neon-js/pull/161/files#diff-9d63cae6f0d7a83a4ee5a8c2d4a4f3d3) (no upstream Secure, SameSite=None), 2 in [`client-factory.test.ts`](https://github.com/neondatabase/neon-js/pull/161/files#diff-12b5d2a3a6cf6c1c7c1fa6c1a3a7e7a7) (factory loop) |
| **2 (correctness)** | Exact `NEON_AUTH_SESSION_COOKIE_NAME` matching via `parseSetCookies`; deletion via parsed `maxAge` / `expires`. Also surface `expires` on `ParsedCookie` | [`session/minting.ts`](https://github.com/neondatabase/neon-js/pull/161/files#diff-1b5f9a4b8d8e7c6d5e4f3a2b1c0d9e8f), [`utils/cookies.ts`](https://github.com/neondatabase/neon-js/pull/161/files#diff-a1b2c3d4e5f6a7b8c9d0e1f2a3b4c5d6) | **+4 tests**: `analytics_session_token_ref` false-match, value containing `max-age=0` next to real `Max-Age=3600`, real `Max-Age=0`, past `Expires` |
| **3 (DX)** | Widen `AuthProxyConfig.log` **and** `NeonAuthServerConfig.log` to `ResolvedNeonAuthLogging \| NeonAuthLogger \| undefined`. New internal `resolveLog(log)` in `logger.ts` (undefined → undefined; pre-resolved → passthrough; partial → `resolveNeonAuthLogging({ logger })` at default warn). Symmetric so adapters can forward `log?: NeonAuthLogger` into either entry without TS2322. | [`proxy/handler.ts`](https://github.com/neondatabase/neon-js/pull/161/files), [`client-factory.ts`](https://github.com/neondatabase/neon-js/pull/161/files), [`logger.ts`](https://github.com/neondatabase/neon-js/pull/161/files) | **+3 handler tests** (undefined, pre-resolved passthrough, partial normalization); **+2 type pins** in `public-surface-types.test.ts` for both `AuthProxyConfig['log']` and `NeonAuthServerConfig['log']` |
| **4 (nit)** | Drop `connection`, `transfer-encoding`, `content-length` from `RESPONSE_HEADERS_ALLOWLIST` | [`proxy/response.ts`](https://github.com/neondatabase/neon-js/pull/161/files) line 6–10 | **+1 test**: asserts none of the three forward |
| **5 (docs)** | Step 2 snippet rewritten: maps `config.cookies.*` to flat fields explicitly, reuses one `proxyConfig` for handler + middleware. Inline comment explaining why `...config.cookies` is wrong. | [`BUILDING-AN-ADAPTER.md`](https://github.com/neondatabase/neon-js/pull/161/files) Step 2 | n/a |
| **6 (docs)** | New **Node-style (non-Fetch) frameworks (Express, Fastify)** section before "What the toolkit gives you for free". Covers all six gotchas: absolute URL + `duplex: 'half'`, `getSetCookie()` (with the Expires/comma trap called out), raw body parsers (Express + Fastify), Express 5 named wildcard `/api/auth/*splat`, AsyncLocalStorage binding for the lazy context factory, ESM-only with Node ≥22.12 interop. | [`BUILDING-AN-ADAPTER.md`](https://github.com/neondatabase/neon-js/pull/161/files) | n/a |

### Notes on FIX 3 (the only place I went slightly beyond your suggestion)

I extended the same widening to `NeonAuthServerConfig.log` for symmetry — an adapter that takes a partial `NeonAuthLogger` from its own users will forward it to both `createAuthServer` and `handleAuthProxyRequest`, and TS2322 on either one breaks the same Step-3 doc pattern. Single `resolveLog()` helper in `logger.ts` covers both call sites. Pre-resolved sinks still pass through unchanged so `createNeonAuth`'s existing `wrapWithLevel` gating is preserved (verified by the `'passes a pre-resolved sink through unchanged'` test).

### Note on FIX 6

I wrote the section directly using your bullet outline rather than waiting on your follow-up — if you'd rather replace it with your own copy (you have the lived experience from the actual adapters), happy to swap.

CHANGELOG bumped to **`0.5.1-beta`** with new entries under **Security** (FIX 1), **Fixed** (FIX 2), and **Changed** (FIX 3, FIX 4).

Re-requesting review on this commit.

cc @andrelandgraf

Shridhad:
Heads up: dropped the manual version bumps in `009a949` so `prepare-release.yml` can own them in the two-stage release pipeline. Packages are back on `main`'s versions (`@neondatabase/auth@0.4.2-beta`, `@neondatabase/neon-js@0.6.2-beta`); the other `package.json` changes (new `./server` and `./auth/server` export conditions, `BUILDING-AN-ADAPTER.md` in `files`) are intact. CHANGELOG entries stay under `[Unreleased]` for the release workflow to roll up.

thekauer (approved):
LGTM

packages/auth/src/server/public-surface-types.test.ts
Shridhad: Good catch, fixed in [`337c67a`](https://github.com/neondatabase/neon-js/pull/161/commits/337c67a).

Went with your suggestion (1): dropped the `toHaveLength(24)` tautology entirely. The `import type {...}` block at the top of the file is what actually catches broken exports (via `pnpm typecheck` → `tsc --noEmit` in the `ci` job), so the runtime `it()` block was just misleading. Replaced it with an inline `Existence pin` comment that documents which mechanism does the work and warns future contributors not to re-add the tautology.

Also corrected the file-header docstring — it previously claimed `pnpm test:ci` enforces the shape pins, which per your empirical test is only true when `--typecheck` is enabled. Corrected to say `pnpm typecheck` (the `ci` step) is what enforces it.

Kept your suggestions 2 and 3 (enable vitest `--typecheck` or import from the built `@neondatabase/auth/server` specifier) as a follow-up pointer in the comment — happy to do either in a follow-up PR if you think the extra coverage is worth the wiring.

PR #168, opened 2026-06-10
fix(auth): always GET session in middleware so Server Actions aren't redirected

Shridhad:
## Summary

Fixes a bug where **Next.js Server Actions on a protected route redirect authenticated users to the sign-in page**, while API routes work fine (reported via Discord).

A Server Action is dispatched as a `POST` to the current page URL. When that page matches the auth middleware matcher (e.g. `/account/:path*`), the middleware runs its internal session lookup by **forwarding the triggering request's method**. That meant:

1. The `get-session` cookie-cache fast path was skipped (it only runs for `GET`).
2. The lookup `POST`ed to the upstream `/get-session` endpoint, which only accepts `GET`, returning a non-OK response.
3. The session was treated as `null` → `requiresRedirect` → redirect to `loginUrl`, even for an authenticated user.

GET navigations and API routes issue `GET` requests, so they were unaffected — exactly matching the report.

### Fix
In `processAuthMiddleware`, the internal session lookup is now always a `GET` (cookies/headers preserved), cloning the request only when the incoming method isn't already `GET`. The middleware lookup is a read, so its method should never depend on how the route was reached.

### Why the demo didn't catch it
`nextjs-neon-auth` previously did all writes through API routes (client `fetch`), never exercising a non-GET request to a matched route. This PR adds a **Server Action** note form on the already-protected `/notes` route so the failure mode stays reproducible (and is a candidate for E2E coverage).

## Changes
- `packages/auth/src/server/middleware/processor.ts` — normalize session lookup to `GET`.
- `packages/auth/src/server/middleware/processor.test.ts` — tests for the Server Action (POST) path + GET passthrough.
- `packages/auth/CHANGELOG.md` — `Fixed` entry.
- `examples/nextjs-neon-auth/app/notes/{actions.ts,server-action-note-form.tsx,page.tsx}` — Server Action repro on `/notes`.
- `CLAUDE.md` — learned-memory bullets (durable conventions/facts).

## Test plan
- [x] `pnpm --filter '@neondatabase/auth' test:node -- --run src/server/middleware/processor.test.ts` (17 passed)
- [x] `pnpm --filter '@neondatabase/auth' typecheck`
- [x] `examples/nextjs-neon-auth` `tsc --noEmit` + `eslint`
- [ ] Manual: sign in, open `/notes`, submit **"Add via Server Action"** — note is added, no redirect. Reverting the `processor.ts` change reproduces the redirect to `/auth/sign-in`.


Shridhad:
Headline finding (multi-lens consensus, 4-of-4)

  processor.ts:280 — when you clone the request as GET, you reuse request.headers which carries Content-Type (e.g. multipart/form-data from a Server Action) and Content-Length from the original POST. Gemini called this a blocker (worried about validation/hang); GPT + canonical + persona agreed on the shape but
  calibrated to should-fix because /get-session is GET-only and tolerates the dangling headers today. Real foot-gun if anything upstream starts validating Content-Length vs body-bytes on GET, or if the same normalization gets applied to a non-tolerant endpoint later.

  Fix: strip content-type, content-length, transfer-encoding on the clone — or leave a comment that they're intentionally inert on a body-less GET.

  Second should-fix

  examples/nextjs-neon-auth/app/notes/actions.ts:70 — deleteNoteAction is exported but has no caller. DeleteNoteButton in delete-note-button.tsx still hits the /api/notes/[id] DELETE route. So the delete path isn't actually demonstrated by the Server Action — only addNoteAction exercises the matched-route POST
  that the PR is fixing. Either wire a ServerActionDeleteButton that posts to deleteNoteAction (so the demo also covers a Server Action on the matched route through a different verb), or drop the action.

  Questions

  1. processor.ts:280 — does handleAuthProxyRequest actually want a NextRequest? Constructing a bare Request drops .nextUrl and .cookies (Next-specific bits). If anything downstream reads those, the clone path silently degrades. Worth a sanity check, or alternatively request.clone() then override method via init.
  (gemini)
  2. server-action-note-form.tsx:38 — the form wraps addNoteAction in an inline client async callback (async (fd) => { await addNoteAction(fd); ... }) instead of passing the action directly to <form action={addNoteAction}>. Question: does Next still POST to /notes in this shape, or does the RPC route through
  /_next/... and skip the /notes middleware matcher? If it's the latter, the demo doesn't actually exercise the path the PR is fixing. Worth confirming — passing the action directly would remove the ambiguity. (gpt)

  Nits (5)

  3. processor.test.ts:207 — the non-GET test asserts method + cookie but not that the new Request differs from config.request. Add expect(passedRequest).not.toBe(config.request) to mirror the GET test's toBe and lock the clone-vs-reuse contract.
  4. CHANGELOG.md:30 — "API routes and normal navigations were unaffected because they issue GET" is a bit off — protected API routes can be non-GET too (e.g. POST to /api/foo matched by middleware). Reword: "matched routes that happened to be GET were unaffected; this normalizes any non-GET method to GET for the
  session lookup."
  5. actions.ts:54 — addNoteAction silently returns on empty/whitespace title. No error surfaced, no required/minLength on the input. Easy for a user to think the action ran. Add HTML5 validation or surface an error.
  6. server-action-note-form.tsx:42 — formRef.current?.reset() unconditionally after await erases any new text the user typed into the still-enabled input while the previous submission was pending. (gemini)
  7. server-action-note-form.tsx:40 — inline async wrapper as form action breaks progressive enhancement — form can't submit without JS. Pass addNoteAction directly + useFormState for the prepend-to-list behavior. (gemini, dovetails with Q2)

thekauer (approved):
LGTM!
Works great!

Shridhad:
Thanks for the thorough pass — pushed `be7e3a8`. Rundown:

**Headline (body headers on the GET clone):** good catch. Added a `stripBodyHeaders` helper that drops `Content-Type`/`Content-Length`/`Transfer-Encoding` on the clone, so the body-less GET no longer advertises a length/type it doesn't have. Avoids the foot-gun if this normalization ever lands on a less tolerant endpoint.

**Q1 (does downstream want a NextRequest?):** checked — the proxy + `handleAuthProxyRequest` only touch standard `Request` members (`headers`/`url`/`method`/`body`); nothing reads `.nextUrl` or `request.cookies`. So the plain `Request` clone is fine. `request.clone()` doesn't let you change the method anyway, so `new Request(url, { method, headers })` is the right tool. Left a comment noting this.

**Q2/#7 (inline wrapper vs direct action):** switched to `<form action={addNoteAction}>`. You were right to flag it — passing it directly removes the ambiguity (Server Actions always POST to the page URL = `/notes`, through the matcher), restores progressive enhancement, and lets React auto-reset the uncontrolled input, which also kills the reset-erases-typed-text issue (#6).

**deleteNoteAction:** dropped it. It was dead (DeleteNoteButton still uses the API route), and `addNoteAction` already exercises the matched-route POST the PR fixes.

**Nits:** #3 added `expect(passedRequest).not.toBe(config.request)` plus assertions that the body headers are stripped and cookies preserved. #4 reworded the changelog ("matched routes that happened to be GET were unaffected; normalizes any non-GET to GET"). #5 added `required` on the input.

Tests/typecheck/lint green.

callicles:
I just hit this as well - would love for you to merge.

PR #175, opened 2026-07-05
fix: lazy init CURRENT_TAB_CLIENT_ID for Cloudflare Workers compat

### LINKED ISSUE #72: crypto.randomUUID() called in global scope breaks Cloudflare Workers
# The offending line
https://github.com/neondatabase/neon-js/blob/b00a8a5e67c8022c75a6042330009a7596f08b77/packages/auth/src/core/better-auth-methods.ts#L30

# The problem explaination
https://developers.cloudflare.com/hyperdrive/observability/troubleshooting/#workers-runtime-errors

TrivCodez:
Closes #72

## Problem

`crypto.randomUUID()` is called at module scope in `packages/auth/src/core/better-auth-methods.ts`, which breaks in Cloudflare Workers. Workers rejects any `crypto.*` calls that execute at import time — regardless of which method you use.

## Solution

Replaced the module-scoped `CURRENT_TAB_CLIENT_ID` export with a lazy getter:

```ts
let _currentTabClientId: string | undefined;

export function getCurrentTabClientId(): string {
  if (!_currentTabClientId) _currentTabClientId = crypto.randomUUID();
  return _currentTabClientId;
}
```

The call is deferred until first use, well after Workers has fully initialised its runtime. Uses `crypto.randomUUID()` directly (no custom UUID implementation needed), just lazily. All call sites updated to use `getCurrentTabClientId()`.


NoahJGersh:
I tested a deployment using this method, and CF Workers was not receptive. It complained the same way with `crypto.randomValues()` as it does with `crypto.generateUuid()`. It seems like rather than a method swap, a more fundamental change (like the lazy initialization you suggested in the connected issue) is the way to go.

TrivCodez:
Thanks for testing on Workers! Since `crypto.randomUUID()` runs at module scope, swapping the method doesn't help — still executes too early.

The fix is lazy init:

```ts
let _currentTabClientId: string | undefined;

export function getCurrentTabClientId(): string {
  if (!_currentTabClientId) _currentTabClientId = crypto.randomUUID();
  return _currentTabClientId;
}
```

Call deferred until first use, well after Workers is initialised.

PR #141, opened 2026-05-12
fix(auth): dual-read session_challange and session_challenge cookie names

thekauer:
## Problem

The Neon Auth server has historically set a session-challenge cookie under the misspelled name `session_challange`. The server is now correcting the spelling to `session_challenge` and dual-writes both names during a transition window — see databricks-eng/neon-cloud#6472. This SDK still hard-coded the legacy name; it needs to read either so old and new server builds in the field both work.

## Summary of changes

- Add `NEON_AUTH_LEGACY_SESSION_CHALLANGE_COOKIE_NAME` for the misspelled name.
- Switch `NEON_AUTH_SESSION_CHALLENGE_COOKIE_NAME` value to the correctly-spelled `session_challenge` (the constant identifier was already correctly spelled).
- In `packages/auth/src/server/middleware/oauth.ts`, dual-read inline at both call sites (`needsSessionVerification` and `exchangeOAuthToken`): try the new name first, fall back to legacy via `??`.
- Tests cover legacy fallback (both `needsSessionVerification` and `exchangeOAuthToken`), new-name happy path, and both-present preferring new name.
- `packages/auth/src/server/proxy/response.ts` audited and is name-agnostic — it allowlists `Set-Cookie` as a header type and pass-through copies cookies without inspecting their names. No change needed there. Existing `response.test.ts` fixtures continue to use `session_challange` since they represent old-server output the SDK should still handle.
- Existing `processor.test.ts:65` fixture also kept on `session_challange` for the same reason.

## Verification

Ran in `packages/auth`:

- `vitest run` over the four touched/adjacent test files (`oauth.test.ts`, `processor.test.ts`, `response.test.ts`, `handler.test.ts`) — 60/60 pass, including 4 new tests for dual-read.
- Full `vitest run` suite — 151/151 pass; two pre-existing test-file failures (`src/index.test.ts`, `src/core/adapter-core.test.ts`) are unrelated, caused by workspace deps `@neondatabase/internal` / `@neondatabase/auth-ui` not being built locally (no `pnpm build`), and reproduce on `main`.
- `eslint` on the three changed files — clean.
- `tsc --noEmit` against the three changed files — no errors. Same workspace-dep issues affect typecheck across the package; they are pre-existing.

## Follow-up

Once all production Neon Auth deployments have shipped the dual-write change (databricks-eng/neon-cloud#6472), the legacy constant + fallback can be dropped. Patch release recommended after merge per the original feedback thread.

PR #197, opened 2026-08-06
fix(auth): preserve OAuth state across redirects

Shridhad:
## Summary
- default proxied and locally managed Auth cookies to `SameSite=Lax` while honoring explicit cookie configuration
- preserve non-colliding request query parameters on same-origin login redirects while keeping configured `loginUrl` parameters authoritative
- prevent request query leakage to external login origins and apply exact/subpath boundaries to login and skipped routes
- accept both canonical `session_challenge` and legacy `session_challange` cookies during the server migration

## Test plan
- [x] `pnpm --filter '@neondatabase/auth' test:node -- --run` (183 tests)
- [x] `pnpm exec tsc -p packages/auth/tsconfig.json --noEmit`
- [x] `pnpm exec eslint packages/auth`
- [x] `pnpm exec tsdown --config packages/auth/tsdown.config.ts`


ruslantalpa:
Review of the OAuth-redirect changes (multi-model + manual verification against the branch). Posting as comments — leaving the approve/request-changes call to a human.

## 🔴 Should-fix before merge

**1. Query params leaked to an external login origin — `packages/auth/src/server/middleware/processor.ts` (the `redirect_login` append loop, ~L194-197)**
When `loginUrl` resolves to a different origin, `sameOriginLoginPathname` is correctly set to `null`, but the final loop still **unconditionally** appends every incoming request query param onto the external `resolvedLoginUrl`:

```
for (const [name, value] of requestUrl.searchParams) {
  resolvedLoginUrl.searchParams.append(name, value);
}
```

So a request to `/dashboard?neon_auth_session_verifier=secret&…` redirects to `https://idp.example.com/sign-in?neon_auth_session_verifier=secret&…` — the verifier and any app params are handed to a third-party origin. Before this PR nothing was appended, so this is a regression introduced here (for the external-`loginUrl` case).

Fix: guard the append with the same-origin check that already exists —
```
if (sameOriginLoginPathname !== null) {
  for (const [name, value] of requestUrl.searchParams) { resolvedLoginUrl.searchParams.append(name, value); }
}
```
…or, if external `loginUrl` is never a supported config, say so and reject it explicitly.

**2. Missing test for the external-origin case — `processor.test.ts`**
The new tests cover same-origin preservation, sibling-prefix, and trailing slash, but there is no test asserting request params are **not** copied onto an external `loginUrl`. Please add one with the fix above (assert `redirectUrl.origin === external` and `!searchParams.has('neon_auth_session_verifier')`).

## 🟡 Worth addressing

**3. `shouldProtectRoute` skip-list still uses loose prefix match — `route-protection.ts:12`**
This PR added boundary-aware `isSamePathOrSubpath()` for the login-path checks, but the `skipRoutes` allow-list is still `pathname.startsWith(route)`. If a configured skip route is a prefix of a protected sibling (e.g. skip `/auth` → `/auth-admin` slips through), enforcement is bypassed. Pre-existing, but it's exactly the class you're hardening here — worth reusing `isSamePathOrSubpath` for the skip-list too.

**4. Colliding-param precedence / parameter-pollution — `processor.ts` (same append loop)**
Request params are appended *after* the configured `loginUrl` params. `.get()` consumers see the configured value (safe), but a last-wins parser downstream (login page / IdP) could treat an attacker-supplied `redirect`/`callbackUrl`/`state` as authoritative. Consider skipping keys already present on `loginUrl`. (Related to #1.)

## ✅ Checked and NOT an issue
- **SameSite `strict`→`lax`**: not a CSRF regression — Lax still blocks cross-site POST/XHR and Better Auth state changes are POST. Lax is the correct OAuth-compatible default; changelog is accurate.
- **`isSamePathOrSubpath` normalization** (trailing slash, root `/`, sibling `/sign-in-admin`, encoded slashes) behaves correctly.

Note: the red Vercel checks are external-deploy authorization, not code.

Shridhad:
@ruslantalpa Addressed the review in `1e07d81`:

- request query parameters are no longer copied to external-origin login URLs
- configured `loginUrl` keys now take precedence over colliding request keys
- `skipRoutes` now use exact/slash-subpath boundaries instead of loose prefixes
- added regression coverage for external origins, collisions, sibling paths, trailing slashes, and root matching

Auth tests pass: 175/175; typecheck and lint pass.

ruslantalpa (approved):

Shridhad:
Rebased onto current `main` (`#168` and `#198`) and resolved the changelog/guidance conflicts while preserving Server Action GET normalization and Better Auth 1.6.23.

Also incorporated the dual-read compatibility from #141: OAuth verification now accepts both `session_challenge` and legacy `session_challange`, with fetch-boundary coverage confirming both Neon Auth cookies reach upstream.

Verification: 183/183 Auth tests, typecheck, lint, and Auth build pass.

Shridhad:
Superseded by #203, which used the same commit from a branch in `neondatabase/neon-js` so trusted CI could run. #203 passed the required checks and has been merged.

PR #198, opened 2026-08-06
chore(deps): bump better-auth to 1.6.23 to match neon-auth server

Shridhad:
## Summary
- Bump Better Auth client stack to **1.6.23** to match the neon-auth server upgrade, with full ecosystem alignment via root `pnpm.overrides` (`@better-auth/*`, `@better-fetch/fetch`, `better-call`, `jose`)
- Move `@daveyplate/better-auth-ui` **3.3.9 → 3.4.0** (required after `apiKeyClient` extraction) and simplify the authClient cast once versions are unified
- Also bump `next` → **16.2.11**, `vite` → **7.3.5**, plus `postcss`/`sharp` overrides so Stage 2’s `pnpm audit --audit-level=high` gate (excluding `examples/`/`e2e`) reports **0** blocking advisories

## Test plan
- [x] `pnpm install` / lockfile resolves to better-auth 1.6.23
- [x] `pnpm run build` (all packages)
- [x] `pnpm typecheck`
- [x] `pnpm test:ci` (auth 160, auth-ui 28, neon-js 46)
- [x] Local Stage 2 audit gate: `pnpm audit --audit-level=high --prod` + non-`examples/`/`e2e` path filter → 0 blocking
- [ ] CI green on this PR (unit + e2e)
- [ ] After merge: re-run Stage 2 publish for `neon-js@v0.6.3-beta` (or cut a fresh patch tag if preferred)


ruslantalpa:
Review of the better-auth 1.6.23 bump (multi-model + manual verification against the regenerated `pnpm-lock.yaml`). Posting as comments — leaving the approve/request-changes call to a human. Only doc/hygiene nits; the two scary machine-generated claims are false positives (details below), and no package `version` fields were bumped, so this is convention-compliant.

## 🟡 Should-fix (low)

**1. CHANGELOG claims `jose 6.2.5` "via root `pnpm.overrides`" — but there is no jose override — `packages/auth/CHANGELOG.md`**
`jose 6.2.5` is a direct dep in `packages/auth/package.json`, not a root override. The lockfile still resolves jose `6.1.2` / `6.2.3` / `6.2.5`, and the changed examples (`neon-auth-magic-link-example`, `nextjs-phone-login`) still declare `^6.1.2`. Either reword the note to the real scope (`packages/auth` direct dep only) or add the override + bump the examples and regenerate the lockfile.

**2. `eslint-config-next` left behind while `next`→16.2.11 — the Next example apps**
`next` was bumped to `16.2.11` but `eslint-config-next` is still `16.0.7` (and `16.1.6` in the orgs example). Dev-tooling only, but it's version-skewed rather than aligned. Bump alongside `next`.

## 🔵 Nuance (not blocking)

**3. `as unknown as` → `as` cast — `packages/auth-ui/src/neon-auth-ui-provider.tsx`**
Behaviorally fine, but the new comment slightly overstates it: `packages/auth` and `packages/auth-ui` still resolve distinct virtual-store `better-auth` instances, so pnpm didn't dedupe the types into one instance — the single `as` is an explicit compatibility assertion, not proof of unification. Suggest rewording. Since the `claude-review` check is red, worth confirming `pnpm typecheck` is clean.

## ✅ Checked and confirmed NOT issues (flagged by automated review, disproven against the lockfile)
- **"`better-auth-ui@3.4.0` requires `better-call@2.0.2`, override forces it off-graph"** — false. The lockfile contains only `better-call@1.3.7` (71 occurrences, zero `better-call@2`); 3.4.0's resolved deps use `better-call: 1.3.7` cleanly. No conflict.
- **"audit gate still fails with 24 high findings (drizzle-orm/react-router-dom/shadcn)"** — those are all `examples/` deps, which the PR's gate (`--prod` + non-`examples/`/`e2e` path filter) explicitly excludes. Worth a final confirmation of the gate command, but not a blocker.
- **Release hygiene**: no package `version` fields changed (only `dependencies`/`overrides`/examples) — compliant with the prepare-release convention.

Note: the red `e2e (react-neon-js)` check is the same GitHub-Actions infra outage as elsewhere today ("Failed to resolve action download info. Error: Service Unavailable"), not a code failure — `unit-tests`, `ci`, and `e2e (nextjs-neon-auth)` all pass. Retrigger it.

ruslantalpa (approved):

Shridhad:
Thanks @ruslantalpa — both should-fixes plus the cast-comment nit are addressed in 2f14278.

**1. `jose` changelog inaccuracy.** You're right, there is no `jose` root override. Reworded the entry to scope it to this package's own direct dependency and to be explicit that other workspace roots keep their own range:

> Related pins updated together via root `pnpm.overrides`: … `better-call` `1.3.7`. This package's own `jose` dependency moves to `6.2.5` (direct dep of `@neondatabase/auth`, not a root override, so other workspace roots keep their own `jose` range).

**2. `eslint-config-next` skew.** Bumped to `16.2.11` in all five Next.js examples (`nextjs-neon-auth`, `nextjs-phone-login`, `nextjs-admin-portal`, `neon-auth-magic-link-example`, and `neon-auth-orgs-example`, which was on `16.1.6`), so the lint config matches the `next` override.

**3. Cast comment.** Also fair — I verified it, and the two packages do still resolve distinct virtual-store instances of the same version:

```
packages/auth/node_modules/better-auth    -> better-auth@1.6.23_…_ng7ypzmqsh5raljhw6lkb4cbxu
packages/auth-ui/node_modules/better-auth -> better-auth@1.6.23_…_gzpytsl7jdvypihx4plk7jzaki
```

So the comment no longer claims the types are unified; it now says the single `as` is an explicit compatibility assertion, and that `pnpm.overrides` keeping both on one version is what makes `as` sufficient instead of `as unknown as`.

**Verification after the change:** `pnpm install` is clean with no unmet peer warnings, and build, `typecheck`, `lint`, and `test:ci` (74 tests) all pass. The Stage 2 audit gate is still clear — the 6 remaining `high` advisories (`react-router` ×5, `drizzle-orm`) root exclusively in `examples/`, which the gate filters out.

PR #202, opened 2026-08-07
feat(auth): add first-party Hono adapter (`@neondatabase/auth/hono/server`)

Shridhad:
## Summary

Phase 2 of the toolkit rollout — Phase 1 promoted the framework-agnostic toolkit to `@neondatabase/auth/server` in [#161](https://github.com/neondatabase/neon-js/pull/161); this PR builds the first-party **Hono adapter** on top of it.

- **New public subpath**: `@neondatabase/auth/hono/server` — exports `createNeonAuth`, `honoAuthHandler`, `neonAuthHonoMiddleware`, `createHonoRequestContext`, plus the toolkit's observability + error-classification surface (same shape as `@neondatabase/auth/next/server`).
- **`neon-js` wrapper**: `@neondatabase/neon-js/auth/hono/server` re-exports the above, matching the mandatory monorepo convention (per `CLAUDE.md`).
- **Example app**: `examples/hono-neon-auth/` — minimal `@hono/node-server` app demonstrating the three mandatory wires and a protected route reading the session.

**Stability: `beta`** — mirrors the Next.js adapter's shape and follows the same versioning contract.

## Design decisions

| Decision | Choice | Why |
|---|---|---|
| Factory name | `createNeonAuth` | Matches Next.js adapter; users can port examples 1:1 (resolved via subpath, so no clash) |
| Subpath convention | `./hono/server` only | Hono is server-only; mirrors `./next/server` |
| RequestContext plumbing | `hono/context-storage` (built-in ALS) | Officially recommended by Hono; user just registers `app.use(contextStorage())` once |
| `hono` peer dep | `>=4.0.0`, **optional** | Same pattern as `next` in this package — no forced install for non-Hono users |

## Tests

25 new unit tests, **all passing** (`packages/auth/src/hono/server/index.test.ts`):

- **Config validation** — secret length, `sessionDataTtl`, `sameSite`, missing secret paths
- **RequestContext contract** — `getCookies` / `getHeader` (case-insensitive) / `getOrigin` (Origin → Referer stem → empty) / `getFramework` / `setCookie` (round-tripped through real `app.fetch()`)
- **Handler** — `/api/auth/` prefix-stripping across 3 URL shapes; raw `Request` + config forwarded to `handleAuthProxyRequest`
- **Middleware** — all 3 `MiddlewareResult.action` branches (allow / redirect_oauth / redirect_login, with and without cookies)

[...]

PR #214, opened 2026-08-23
feat(auth): add first-class Nuxt integration

mezotv:
## Summary
- add a Vue-backed `@neondatabase/auth/nuxt` client
- add an H3/Nitro server adapter with event-bound server methods, an auth proxy and opt-in route protection
- add a Nuxt UI example with auth flows, organizations, notes and shared E2E coverage
- export the Nuxt adapter from `@neondatabase/neon-js`
- prevent Vue clients from entering the React auth UI provider

## Test plan
- [x] Build the SDK packages and Nuxt example
- [x] Typecheck `@neondatabase/auth`, `@neondatabase/auth-ui`, `@neondatabase/neon-js` and the Nuxt example
- [x] Run the auth unit tests (252 passing)
- [x] Lint the auth, auth-ui, neon-js and Nuxt example packages


packages/auth/src/nuxt/server/middleware.ts
```diff
@@ -0,0 +1,147 @@
...
+import {
+  DEFAULT_AUTH_SKIP_ROUTES,
+  processAuthMiddleware,
+  type MiddlewareResult,
+  type ResolvedNeonAuthLogging,
+  type SessionCookieSameSite,
+} from '../../server';
+import { createFetchRequest } from './handler';
+
+const DEFAULT_PROTECTED_ROUTES = ['/'] as const;
```
Shridhad: This is unsafe for a global Nitro middleware default. `['/']` matches `/_nuxt/*`, public assets, and unrelated server/API routes; unauthenticated requests to those paths can be redirected to the login page. That can prevent the sign-in page itself from loading its JS/CSS. Hono PR #202 avoids this by scoping middleware to a protected route. Could we require explicit protected prefixes, default to none, or add/document framework asset exclusions? The README should show the safe wiring either way.

packages/auth/src/nuxt/server/index.test.ts
```diff
@@ -0,0 +1,298 @@
...
+    expect(upstreamBody).toBe(rawBody);
+    expect(response.status).toBe(201);
+    expect(getHeaders(response, 'set-cookie')).toHaveLength(2);
+    expect(await new Response(response.body as BodyInit).text()).toBe(
+      '{"ok":true}'
+    );
+  });
+});
+
+describe('Nuxt middleware', () => {
```
Shridhad, line 213: These tests cover the matcher and result mapping, but almost never exercise `auth.middleware()` end-to-end—the one real invocation is the OAuth verifier case below. Please add wiring tests for: (1) a public route outside `protectedRoutes` doing no fetch/redirect, (2) an unauthenticated protected route redirecting to login, (3) the login route being allowed without a loop, and (4) `/api/auth/*` staying allowed. Those paths validate the Nuxt-specific early return and `skipRoutes` composition rather than re-testing toolkit internals.

packages/auth/README.md
```diff
@@ -267,15 +267,45 @@ See the [Next.js Setup Guide](./NEXT-JS.md) for comprehensive documentation incl
...
+  cookies: {
+    secret: process.env.NUXT_NEON_AUTH_COOKIE_SECRET!,
+  },
+});
+```
+
+Mount `auth.handler()` at `server/api/auth/[...path].ts` and use
+`auth.middleware()` from Nitro middleware. Bind server calls to the current
+request with `auth.withEvent(event)`. See
+[`examples/nuxt-neon-auth`](../../examples/nuxt-neon-auth) for a working Nuxt
```
Shridhad, line 331: This example is not present in this PR, so the published README will contain a broken link and adopters will not have the three concrete Nitro files needed to wire the handler and middleware. Please either include the stacked example here (as Hono #202 does) or remove this link and add complete `server/api/auth/[...path].ts` and `server/middleware/*.ts` snippets, including a safe `protectedRoutes` configuration.

packages/auth/src/nuxt/server/index.ts
```diff
@@ -0,0 +1,97 @@
...
+  const proxyConfig = {
+    baseUrl,
+    cookieSecret: cookies.secret,
+    sessionDataTtl: cookies.sessionDataTtl,
+    domain: cookies.domain,
+    sameSite: cookies.sameSite,
+    log,
+  };
+
+  return {
```
Shridhad, line 37: Can we confirm this API divergence is necessary? The extracted toolkit and Hono #202 keep `createNeonAuth()` consistent with Next by returning the `NeonAuthServer` proxy and attaching `.handler()` / `.middleware()`. Here every server call becomes `auth.withEvent(event).getSession()`. Explicit event binding is concurrency-safe, but it makes Nuxt a different template for future adapters. If H3/Nitro request context cannot safely provide the same surface, please document that constraint and why `withEvent` is the intentional framework-specific exception.

Shridhad (changes requested):
The Nuxt implementation does reuse the extracted `@neondatabase/auth/server` toolkit correctly: `RequestContext`, `createAuthServer`, `handleAuthProxyRequest`, `processAuthMiddleware`, validation, and logging remain shared rather than being copied. The H3 bridge also handles raw bodies and separate `Set-Cookie` values carefully, and the explicit event binding has a good concurrency test.

The main blocker is the route-protection default. Unlike Hono PR #202, where middleware is scoped to a protected route, Nitro middleware is global and this defaults to protecting every path. That can redirect Nuxt assets and nominally public server routes. The docs currently encourage using `auth.middleware()` without showing a safe `protectedRoutes` configuration, and the missing example means there is no integration artifact demonstrating the intended wiring.

I also think we should make an explicit API decision before calling Nuxt a reference adapter: Next and Hono return the `NeonAuthServer` proxy directly and attach `.handler()` / `.middleware()`, while Nuxt returns `{ withEvent, handler, middleware }`. Explicit event binding is safe, but it means future framework adapters cannot follow one consistent application-facing pattern. If Nitro request context cannot safely support the same surface, documenting this as a deliberate Nuxt-specific exception is reasonable.

Requested changes: make the middleware default safe (or explicitly exclude framework assets and document whole-app protection), add real middleware wiring tests, and either include the referenced Nuxt example or remove the broken link and provide complete setup snippets.

packages/auth/src/nuxt/server/middleware.ts
mezotv: Changed the default protected route list to empty. Global Nitro middleware now only handles OAuth verifier callbacks unless the app passes explicit protected prefixes. The README shows the safe configuration.

packages/auth/src/nuxt/server/index.test.ts
mezotv: Added end-to-end middleware wiring tests for the safe public default, an unauthenticated protected route, the login route and the auth API route.

packages/auth/README.md
mezotv: The Nuxt example is now included in this PR. The README also includes the route handler, global middleware with explicit protected prefixes and an event-bound server method example.

packages/auth/src/nuxt/server/index.ts
mezotv: Documented this as a Nuxt-specific API choice. A published adapter cannot depend on Nuxt application auto-imports or Node-only request storage across every Nitro preset, so withEvent keeps request state explicit and portable.

packages/auth/src/nuxt/server/middleware.ts
Shridhad: Looks good. Empty default plus explicit prefixes in the example/README is the right Nitro default.

packages/auth/src/nuxt/server/index.test.ts
Shridhad: Thanks — the new public-default, protected-redirect, login-skip, and `/api/auth` wiring tests cover the Nuxt-specific paths I was missing.

packages/auth/README.md
Shridhad: Resolved. The example is in-tree, E2E covers it, and the README now has the handler/middleware/`withEvent` snippets.

packages/auth/src/nuxt/server/index.ts
Shridhad: That rationale is enough for me. Documenting `withEvent` as the Nuxt-specific exception (rather than a new adapter template) is the important part — thanks.

packages/auth/README.md
```diff
@@ -267,15 +267,79 @@ See the [Next.js Setup Guide](./NEXT-JS.md) for comprehensive documentation incl
...
+export default auth.handler();
+```
+
+```typescript
+// server/middleware/auth.ts
+import { auth } from '../utils/auth';
+
+export default auth.middleware({
+  loginUrl: '/auth/sign-in',
+  protectedRoutes: ['/account', '/dashboard', '/notes'],
```
Shridhad: Nit: this snippet still protects `/dashboard` and omits `/organization`, while `examples/nuxt-neon-auth/server/middleware/auth.ts` uses `['/account', '/organization', '/notes']` and leaves `/dashboard` to a client `watchEffect`. Worth aligning so the published docs match the example adopters will copy.

e2e/tests/organizations.spec.ts
```diff
@@ -43,13 +43,5 @@ test.describe('Organizations', () => {
...
     await expect(page).toHaveURL('/organization/members');
-
-    // Navigate to teams page
-    await page.goto('/organization/teams');
-    await expect(page).toHaveURL('/organization/teams');
-
-    // Navigate to API keys page
-    await page.goto('/organization/api-keys');
-    await expect(page).toHaveURL('/organization/api-keys');
   });
```
Shridhad, line 48: Nit: dropping `/organization/teams` and `/organization/api-keys` here also reduces Next.js coverage, since this spec is shared. If Nuxt does not ship those views, consider gating them with `test.skip(targetApp === 'nuxt-neon-auth', ...)` instead of removing them for every app.

Shridhad (approved):
Follow-up looks good. The previous blockers are addressed: route protection is opt-in, middleware wiring tests cover the Nuxt-specific paths, the example is in-tree with handler/middleware/`withEvent` docs, and `withEvent` is documented as a Nuxt-specific exception rather than a new adapter template. Unit tests plus React/Next/Nuxt E2E are green.

The adapter still sits on the shared toolkit correctly. Two leftover nits below, neither blocking.

packages/auth/README.md
mezotv: Aligned the README with the example: `/account`, `/organization` and `/notes` are now the documented protected prefixes.

e2e/tests/organizations.spec.ts
mezotv: Restored Teams and API Keys navigation coverage for Next.js in a separate suite that is skipped only for the Nuxt target.

PR #217, opened 2026-09-05
fix(auth): force GET method and undefined body for get-session path in handleAuthRequest (#200)

### LINKED ISSUE #200: @neondatabase/auth: neonAuthMiddleware proxies caller's HTTP method to /get-session -> 405 on any non-GET
## Summary

`@neondatabase/auth`'s Next.js middleware, when it forwards the request to the auth service's `/get-session` endpoint, reuses the caller's HTTP method. Any non-GET request to a middleware-gated path (say `PATCH /api/v1/foo`) becomes a `PATCH /get-session` at the auth service, which answers 405. The middleware then treats the null response as an absent session and redirects to the sign-in page.

Net effect: **every state-mutating request from a signed-in user gets kicked to sign-in**, even though the browser's session cookie is perfectly valid.

## Environment

- `@neondatabase/auth@0.4.2-beta`
- Next.js 16.3.0 (Turbopack)
- Node 24

## Reproduce

Any signed-in user hitting a non-GET route gated by the middleware:

```ts
// A protected API route
export async function PATCH(request: NextRequest) { ... }
```

With the recommended `neonAuthMiddleware` in place (or `auth.middleware()` invoked inside your own middleware), the request never reaches the handler — it 401s or redirects to the sign-in page, regardless of the session cookie.

## Where in the shipped bundle

`node_modules/@neondatabase/auth/dist/next/server/index.mjs`, function `handleAuthRequest` (line ~1072 in the current beta). The inner fetch to the auth service:

```js
const handleAuthRequest = async (baseUrl, request, path, log) => {
    ...
    method: request.method,   // <-- forwards the caller's method verbatim
    ...
};
```

`/get-session` (and every other endpoint the middleware might reach) is GET-only on the auth service.

## Proposed fix

Two options:

1. **Use the endpoint's own method constant** — each auth-service endpoint has a known HTTP verb (`/get-session` = GET). Route based on `path`, not the caller's method.
2. **Early-return for non-GET at the middleware level** — for POST/PATCH/PUT/DELETE, don't call `/get-session` at all; just verify the cookie signature locally and let the request through. The rolling Set-Cookie refresh only needs to happen on GETs (which is what browser navigation is anyway), so non-GET requests can skip the refresh without hurting session longevity in practice.

Happy to send a PR if either shape is preferred.

## Our current workaround

We skip the `auth.middleware()` delegation for non-GET and fall back to plain `auth.getSession()` (which internally always calls the auth service with GET). Downside: the rolling Set-Cookie refresh only runs on GETs, but in normal use that's fine — page loads and drawer fetches are GETs, cookies rotate on time.

Reference in our codebase: `src/proxy.ts` (branch `if (request.method !== 'GET')`).

## Impact

Silent kicks whenever a signed-in user submits a form, drags a card, or triggers any state-mutating request. Reads as "my session keeps expiring" from a user perspective; the fingerprint is `hasSessionCookie: true` with the request method being non-GET.

Tyagiquamar:
Fixes #200.

In @neondatabase/auth, when handleAuthRequest proxied requests to the auth server's get-session path, it forwarded request.method verbatim. When state-mutating requests (such as POST, PUT, PATCH, DELETE API calls or Server Actions) were issued on middleware-gated routes by a signed-in user, handleAuthRequest attempted a POST /get-session request to the upstream auth server, which responded with 405 Method Not Allowed. The middleware interpreted the 405 as an absent session and redirected signed-in users to sign-in.

This PR updates handleAuthRequest in packages/auth/src/server/proxy/request.ts to explicitly use GET method and undefined body when path === 'get-session'.

### DIFF
--- a/packages/auth/src/server/proxy/request.test.ts
+++ b/packages/auth/src/server/proxy/request.test.ts
@@ -31,4 +31,23 @@ describe('handleAuthRequest cookie forwarding', () => {
     const upstreamHeaders = new Headers(requestInit?.headers);
     expect(upstreamHeaders.get('cookie')).toBe(`${legacyCookie}; ${canonicalCookie}`);
   });
+
+  test('forces GET method for get-session path even when caller request method is POST', async () => {
+    const fetchSpy = vi
+      .spyOn(globalThis, 'fetch')
+      .mockResolvedValue(new Response(null, { status: 200 }));
+    const request = new Request('https://app.example.com/api/mutate', {
+      method: 'POST',
+      body: JSON.stringify({ data: 'test' }),
+      headers: {
+        'Content-Type': 'application/json',
+      },
+    });
+
+    await handleAuthRequest('https://auth.example.com', request, 'get-session');
+
+    const requestInit = fetchSpy.mock.calls[0]?.[1];
+    expect(requestInit?.method).toBe('GET');
+    expect(requestInit?.body).toBeUndefined();
+  });
 });
--- a/packages/auth/src/server/proxy/request.ts
+++ b/packages/auth/src/server/proxy/request.ts
@@ -34,12 +34,14 @@ export const handleAuthRequest = async (
 	log?: ResolvedNeonAuthLogging,
 ) => {
 	const headers = prepareRequestHeaders(request);
-	const body = await parseRequestBody(request);
+	const isGetSession = path === 'get-session';
+	const method = isGetSession ? 'GET' : request.method;
+	const body = isGetSession ? undefined : await parseRequestBody(request);
 
 	try {
 		const upstreamURL = getUpstreamURL(baseUrl, path, { originalUrl: new URL(request.url) });
 		const response = await fetch(upstreamURL.toString(), {
-			method: request.method,
+			method,
 			headers: headers,
 			body: body,
 		});


Shridhad:
Thanks for the PR — the user-facing bug this targets was already fixed in https://github.com/neondatabase/neon-js/pull/168.

`processAuthMiddleware` always issues the internal session lookup as a GET (cookies preserved, body-framing headers stripped), so a Server Action / PATCH / POST on a protected route no longer gets forwarded to upstream `/get-session` as a non-GET. That landed on `main` on Aug 6 and shipped in `@neondatabase/auth@0.5.0-beta`.

#200 was filed against `0.4.2-beta`, which predates that fix. The `handleAuthRequest` path this PR changes is no longer on the middleware hot path for this failure mode.

@Tyagiquamar if you're still seeing signed-in users redirected on non-GET protected routes after upgrading to `0.5.0-beta` (or current `main`), please share the package version and a short repro and we can take another look. Otherwise I think we can close this in favor of #168.

Tyagiquamar:
Thanks @Shridhad. Closing this in favor of #168, which already landed the GET session lookup (cookies preserved, body-framing headers stripped) and shipped in `@neondatabase/auth@0.5.0-beta`. Appreciate the pointer.

PR #218, opened 2026-09-08
fix(auth): skip Set-Cookie write-back in Next.js RSC render phase

### LINKED ISSUE #196: next: getSession() slow path calls cookies().set() during Server Component render — "Cookies can only be modified in a Server Action or Route Handler"
**Package:** `@neondatabase/auth@0.4.2-beta` (Next adapter, `@neondatabase/auth/next/server`)
**Next.js:** 16.2.12 (App Router)

## Summary

`auth.getSession()` called from a React Server Component crashes the render with

```
Error: Cookies can only be modified in a Server Action or Route Handler.
```

whenever the `session_data` cache cookie is missing or expired (default `sessionDataTtl` is 300 s, so: sign in, idle >5 minutes, load any page that calls `getSession()` in an RSC).

## Root cause

`getSession()` has two paths:

- **Fast path:** a valid signed `session_data` cache cookie → returns the cached payload, no cookie writes. This is why pages work right after sign-in.
- **Slow path:** cache cookie missing/expired → the SDK fetches `get-session` upstream, the upstream responds with refreshed `Set-Cookie` headers, and the SDK writes them back unguarded. In `dist/next/server/index.mjs` that is the `Set-Cookie` loop in `fetchWithAuth` (`await ctx.setCookie(...)`, ~line 844), which lands in `createNextRequestContext`'s `setCookie` → `cookies().set(...)` (~line 1012).

Next.js forbids cookie writes during Server Component render, so the slow path throws whenever it runs inside an RSC.

## Repro

1. `createNeonAuth(...)` per the docs; sign in.
2. Delete the `session_data` cookie in devtools (or wait past `sessionDataTtl`) while keeping the session-token cookie.
3. Load a page whose RSC does `await auth.getSession()`.
4. Render throws `Cookies can only be modified in a Server Action or Route Handler`.

## Expected

`getSession()` is documented as usable in Server Components ("Where to use: React Server Components …" in the `createNeonAuth` JSDoc), so the slow path should not attempt cookie writes there — e.g. guard `ctx.setCookie` (swallow/queue when writes are illegal), or skip the write-back when not in a Server Action / Route Handler context.

## Workaround we ship

- Run `auth.middleware()` on all gated routes so the refresh happens in middleware, where response-cookie writes are legal.
- For public pages that can't sit behind the middleware (it hard-redirects sessionless visitors to `loginUrl`, and there's no `skipRoutes` config), do a passive read: `auth.getSession({ query: { disableRefresh: "true" } })` — the upstream then sends no `Set-Cookie`, so the write-back loop never runs.

Related: #195 — the middleware itself mis-handles Server Action POSTs, which is why our middleware matcher needs a `missing: next-action` skip.

Tyagiquamar:
## Summary
- Fixes #196: `auth.getSession()` slow path no longer calls `cookies().set()` during React Server Component render.
- Adds optional `RequestContext.canSetCookies` so the server toolkit can return refreshed session data while skipping cookie write-back in read-only contexts.
- Next.js adapter implements `canSetCookies` via `areCookiesMutableInCurrentPhase` (writes still allowed in Route Handlers and Server Actions).

## Root cause
When `session_data` is missing or expired, `getSession()` fetches upstream `get-session`, receives `Set-Cookie`, and `fetchWithAuth` unconditionally forwarded them through `ctx.setCookie()` → `cookies().set()`. Next.js forbids that during RSC render.

## Test plan
- [x] `cd packages/auth && node --run test:ci` (249 tests, including new `adapter.test.ts`)
- [x] RED reproduced on `upstream/main` before fix (test B throws `ReadonlyRequestCookiesError`)
- [x] `node --run lint` in `packages/auth`
- [x] RSC fast path (valid cache): no cookie writes
- [x] RSC slow path (expired/missing cache): session returned, no illegal writes
- [x] Route Handler / Server Action phase: cookie writes preserved

### DIFF
--- a/packages/auth/src/next/server/adapter.test.ts
+++ b/packages/auth/src/next/server/adapter.test.ts
@@ -1,0 +1,321 @@
+import { afterEach, beforeEach, describe, expect, test, vi } from 'vitest';
+import { SignJWT } from 'jose';
+import { createAuthServer } from '@/server/client-factory';
+import { createNextRequestContext } from './adapter';
+import {
+  NEON_AUTH_SESSION_COOKIE_NAME,
+  NEON_AUTH_SESSION_DATA_COOKIE_NAME,
+} from '@/server/constants';
+import type { RequireSessionData } from '@/server/types';
+import type { BetterAuthSession, BetterAuthUser } from '@/core/better-auth-types';
+
+const TEST_SECRET = 'test-secret-at-least-32-characters-long!';
+const TEST_BASE_URL = 'https://auth.example.com';
+
+class ReadonlyRequestCookiesError extends Error {
+  constructor() {
+    super(
+      'Cookies can only be modified in a Server Action or Route Handler.'
+    );
+    this.name = 'ReadonlyRequestCookiesError';
+  }
+}
+
+const sessionPayload: RequireSessionData = {
+  session: {
+    id: 'session-123',
+    userId: 'user-123',
+    token: 'opaque-token',
+    expiresAt: new Date(Date.now() + 3_600_000),
+    createdAt: new Date(),
+    updatedAt: new Date(),
+    ipAddress: '127.0.0.1',
+    userAgent: 'test-agent',
+  } as BetterAuthSession,
+  user: {
+    id: 'user-123',
+    email: 'test@example.com',
+    emailVerified: true,
+    name: 'Test User',
+    image: null,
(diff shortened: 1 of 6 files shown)


Tyagiquamar:
Hi, friendly ping for review on this PR when you have a moment. Happy to address any feedback. Thanks!

